PR #3, opened 2020-12-02
Syncing Upstream to Downstream CVO enablement

zvonkok:
This PR labels the Dockefile for CVO inclusion.
Adds auto-generation of CVO manifests and adds image-references and ClusterOperator CR.


zvonkok:
@dagrayvid @ArangoGutierrez Sync from upstream, all future changes regarding CVO should go here.

dagrayvid:
@zvonkok can we delete the cvo-manifests directory that is still there? (manifests/ is also there)

PR #5, opened 2020-12-03
CVO enhancements

zvonkok:
Added correct naming of resources.
Added CVO annotations.


jmencak:
One thing that will definitely need to be added in the future is operator e2e and upgrade tests for this repo.

scripts/rename.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash 
+
+set -e 
+
+for i  in $(ls [0-9][0-9][0-9][0-9].yaml)
```
jmencak, line 5: Nits:
1) s/for i  in/for i in/
2) do we need the semicolons at the end of variable assignments?
3) please remove trailing space
4) I understand this is a quick hack, but I'd like to see something like `yq` for this in the future.

jmencak (commented):
Only partial testing done.  Looks good on the whole, just a few nits.  Thanks for the PR, it is a definite improvement over the current state.

scripts/rename.sh
zvonkok: Was thinking of `yq` but this is another tool that needs to be installed and bash works everywhere ... But good point.

jmencak: Understood.  NTO didn't use to use `yq` until it was needed as a workaround to patch (badly) generated yamls.  And NTO is not the only operator using `yq` for patching manifests, so I wouldn't be too worried about using a slightly non-standard tool.

zvonkok: Ok thats great, will update the PR.

zvonkok: Ok changed to yq.

jmencak: OK, not sure I should take back what I said about `yq`, but you don't seem to be using the `yq` https://github.com/openshift/build-machinery-go/ is using.  That project is using https://github.com/mikefarah/yq/  and this is what seems to be used and recommended.  As for the rest, it would be nice to see the extra spaces and ';' gone unless they're really needed.

scripts/rename.sh
```diff
@@ -0,0 +1,11 @@
...
+do
+  KIND=$(yq .kind $i | tr -d '"');
+  NAME=$(yq .metadata.name $i | tr -d '"');
```
dagrayvid, line 8: @zvonkok the PR mostly looks good to me aside from the nits Jiri mentioned. I tried using the yq from https://github.com/mikefarah/yq/ and the following changes were needed. (There are several ways to install it in the README, not sure if you want to add one to the Makefile to replace the pip install)

```suggestion
  KIND=$(yq r $i kind | tr -d '"');
  NAME=$(yq r $i metadata.name | tr -d '"');
```

zvonkok: Reverted to bash. Lets keep this in mind if we need sophisticated yaml manipulation/reading.

Dockerfile
```diff
@@ -30,7 +30,9 @@ COPY --from=builder /workspace/manager .
 COPY config/recipes/ /opt/sro/recipes/
 COPY manifests /manifests
 
-USER nonroot:nonroot
+RUN useradd -r -u 499 nonroot
+RUN groupadd   -g 499 nonroot
```
dagrayvid, line 34: This line causes `make local-image-build` to fail for me, because "groupadd: group 'nonroot' already exists" from the useradd command.

```suggestion
RUN groupadd   -g 499 nonroot
```

dagrayvid:
Thanks for the PR. Other than looking into re-adding the nonroot group later (which seems to be added automatically by the useradd on my machine, at least...) this looks good.

/lgtm

PR #6, opened 2020-12-04
Default GOOS and GOARCH to current "go env" value

andymcc:
The Dockerfile defaults to linux/amd64, this PR sets the GOOS and GOARCH
env values rather than forcing linux/amd64.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@ COPY vendor/ vendor/
 
 
 # Build
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -mod=vendor -a -o manager main.go
+RUN CGO_ENABLED=0 GO111MODULE=on go build -mod=vendor -a -o manager main.go
 
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details


yselkowitz:
better just not to set those variables entirely.

ArangoGutierrez (changes requested):
Why not just removing GOOS and GOARCH from line 22, the GO compiler will read the values from the ENV

ArangoGutierrez:
> better just not to set those variables entirely.

Agree
/hold

ArangoGutierrez:
For reference this is the output of `go env` on the base image
```bash
[eduardo@fedora-ws node-feature-discovery-operator]$ podman run -it --rm --entrypoint /bin/bash registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.7
[root@e74a2ac1158b origin]# go env
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/go/.cache"
GOENV="/root/.config/go/env"
GOEXE=""
GOFLAGS="-mod=vendor"
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GOMODCACHE="/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/lib/golang"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/lib/golang/pkg/tool/linux_amd64"
GCCGO="gccgo"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build062095506=/tmp/go-build -gno-record-gcc-switches"
```

andymcc:
That is a better idea, thanks - PR should be changed now.

ArangoGutierrez (approved):
Thanks for looking at this
/lgtm

PR #9, opened 2020-12-07
Removed abberviation in namespace and added preamble for SRO to report conditions early

zvonkok:
Removed abberviation in namespace and added preamble for SRO to report conditions early

### DIFF
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -41,4 +41,5 @@ resources:
   - ../manager
   - ../prometheus
   - ../clusteroperator
+  - ../cr
 namespace: openshift-sro
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 SPECIALRESOURCE  ?= driver-container-base
-NAMESPACE        ?= openshift-sro
+NAMESPACE        ?= openshift-special-resource-operator
 PULLPOLICY       ?= IfNotPresent
 REGISTRY         ?= quay.io
 ORG              ?= openshift-psap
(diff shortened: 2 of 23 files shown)


config/default/kustomization.yaml
```diff
@@ -41,4 +41,5 @@ resources:
   - ../manager
   - ../prometheus
   - ../clusteroperator
+  - ../cr
 namespace: openshift-sro
```
ArangoGutierrez, line 45: Shouldn't this ns also be changed?

ArangoGutierrez (commented):
Looks good, just a minor comment

config/default/kustomization.yaml
zvonkok: This is a runtime settings, changed by kustomize when deployed via make deploy. Kustomize will read the setting from the Makfile. 

For CVO the important manifests are in /manifests

ArangoGutierrez: Yeah that is what I was thinking, but wanted to ask, I still don't fully understand Kustomize yet 
/lgtm

dagrayvid:
Overall it looks good, but can you explain (maybe in a comment in the CR manifest) the reason for having the preamble be a specialresource?

zvonkok:
@dagrayvid This is to kickoff the reconcile loop. Only in the reconcile loop we have all tools to update cluster resources. All the other execution is happening in some vendor libraries.

dagrayvid:
Thanks for the response.

/lgtm

PR #15, opened 2021-06-25
Sync downstream for 4.9

dagrayvid:
This PR will bring in all the changes from upstream into downstream so we can begin ART builds based on openshift/special-resource-operator for 4.9.

The only additional commit removes the github actions .workflow files and adds them to .gitignore.


dagrayvid:
This make verify error is because openshift/release config is set to go1.15. Updating in separate PR....

PR #33, opened 2021-09-01
Bug 2000226: Unable to have multiple charts in one configmap

dagrayvid:
This PR syncs the downstream repo with upstream, most importantly to fix the referenced BZ.

The other minor changes from upstream:
- Make sure debug pods in e2e test are assigned to Ready nodes only.
- Add olm skiprange to the "master copy" of the CSV config/manifests/bases/special-resource-operator.clusterserviceversion.yaml so that future bundle updates will keep the change from #32 
- Add pacevedo to reviewers
- Add upstream fix to infoscale special resource.

/cc @SchSeba


wabouhamad (commented):
/lgtm

PR #36, opened 2021-09-30
Fix ReplicaSet upgrade checks

pacevedom:
In the event of upgrades to deployments, old replica sets were stuck checking a condition that wont ever be true: available replicas == desired replicas. When a RS is set to 0 replicas, the availableReplicas field from the status disappears, rendering an always-false condition to check if the reconcile loop is finished for a resource.
There was also an additional bug that prevented checks for more than the first RS.

Signed-off-by: Pablo Acevedo Montserrat <pacevedo@redhat.com>

### DIFF
--- a/pkg/poll/poll.go
+++ b/pkg/poll/poll.go
@@ -223,24 +223,27 @@ func ForDeployment(obj *unstructured.Unstructured) error {
 				return false
 			}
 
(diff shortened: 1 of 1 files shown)


pacevedom:
Closing in favor of https://github.com/openshift/special-resource-operator/pull/38

PR #37, opened 2021-09-30
Add preamble creation to e2e test

pacevedom:
Available test, which is the first one to run, is based on the existence of ClusterOperator resource. This does not exist unless any of these conditions are met:

- Deployed manually from git.
- A SpecialResource is in reconcile loop.

Deploying from OLM means none of these conditions are met, therefore we force the creation of preamble SpecialResource to trigger the creation of ClusterOperator.

Signed-off-by: Pablo Acevedo Montserrat <pacevedo@redhat.com>

### DIFF
--- a/test/e2e/basic/available.go
+++ b/test/e2e/basic/available.go
@@ -3,18 +3,40 @@ package e2e
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pacevedom:
Closing in favor of https://github.com/openshift/special-resource-operator/pull/38

PR #40, opened 2021-10-01
Bug 2009888: Rename project to openshift-special-resource-operator 

dagrayvid:
The SRO CSV name collides with the community operator name.  Because it's easier to rename the one that is not yet publicly available, we can fix this by renaming the downstream CSV (and project) to openshift-special-resource-operator. See BZ and discussion in the ART slack channel for why this is needed.

This will mean that going forward the upstream and downstream bundle will diverge. The upstream bundle can become the community version, and downstream can be the ART version. When we cherry-pick changes from upstream, we will need to regenerate the downstream bundle.

/cc @bthurber @pacevedom @thiagoalessio


wabouhamad (commented):
/lgtm

PR #43, opened 2021-10-05
Bug 2010911: RenderOperatingSystem() returns wrong OS version on OCP 4.7.24

pacevedom:
Fix RHEL version resolution under OCP 4.7.24+, where RHEL version changed from 8.3 to 8.4 and this was mistakenly fixed to 8.3 for all OCP 4.7.z versions.


dagrayvid:
/retitle Bug 2010911: RenderOperatingSystem() returns wrong OS version on OCP 4.7.24

dagrayvid:
/hold

We should apply these changes to master first, right?

PR #58, opened 2021-10-14
Bug 1996916: Special Resource Operator(SRO) - Fail to deploy simple-kmod on GCP

pacevedom:
When deploying in GCP the crane package gets into an infinite loop if traffic towards metadata server is dropped. This makes SRO unresponsive.
Changing authn discovery from crane package into a manual approach where we grab the pull secrets from a secret will avoid the use of the metadata server while still being able to authenticate and obtain the same info.


pacevedom:
cc @dagrayvid

PR #67, opened 2021-11-08
Update README url and maintainers

bthurber:


dagrayvid:
Thanks @bthurber.

For changes to the CSV we need to modify the file `config/manifests/bases/openshift-special-resource-operator.clusterserviceversion.yaml` which is a "skeleton" of the CSV. Otherwise future executions of `make update-bundle` or `make bundle` will overwrite the changes. 

I would recommend making the change to the "skeleton" and then update the bundle with make. Since we want this change backported to 4.9 we should probably get this merged before #68

/cc @pacevedom

bthurber:
> Thanks @bthurber.
> 
> For changes to the CSV we need to modify the file `config/manifests/bases/openshift-special-resource-operator.clusterserviceversion.yaml` which is a "skeleton" of the CSV. Otherwise future executions of `make update-bundle` or `make bundle` will overwrite the changes.
> 
> I would recommend making the change to the "skeleton" and then update the bundle with make. Since we want this change backported to 4.9 we should probably get this merged before #68
> 
> /cc @pacevedom

ack.  Will close this PR and submit a new PR.

PR #73, opened 2021-11-12
Bug 2019096: Update SRO leader election timeout to support SNO

qbarrand:
Use library-go's default lease duration.
Refactor command-line parsing and add unit tests.


main.go
```diff
@@ -49,22 +48,21 @@ func init() {
...
 
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		Scheme:             scheme,
-		MetricsBindAddress: metricsAddr,
+		MetricsBindAddress: cl.MetricsAddr,
 		Port:               9443,
-		LeaderElection:     enableLeaderElection,
+		LeaderElection:     cl.EnableLeaderElection,
 		LeaderElectionID:   "b6ae617b.openshift.io",
+		LeaseDuration:      &cl.LeaderElectionLeaseDuration,
```
mhrivnak, line 65: I think you'll need to set `RenewDeadline` also. It defaults to 10s, after which the controller [gives up its role as leader and releases its lease](https://github.com/kubernetes/client-go/blob/608b521/tools/leaderelection/leaderelection.go#L281-L290).

The default `RetryPeriod` of 2s also results in a lot of API server traffic, so it's probably wise to use the openshift default for that (26s).

mhrivnak (changes requested):
Hi, thanks for humoring my review. I'm just starting to get to know SRO and figure some code review is one good way to do so. I also have a lot of background with leader election, so this one caught my eye.

main.go
mhrivnak: @liqcui since you LGTM'd, I'll just elaborate on my above comment and say I don't think this fixes the reported bug, because while the API service is unavailable, the controller will still give up its lease after only 10s. The change as-is will only prevent other instances from forcibly becoming the leader. Assuming we want continuity such that the same controller instance can be the leader both before and after an API service period of unavailability, then the `RenewDeadline` value needs to also be changed.

liqcui: Sorry, I'm not fully aware of this point, I will remove my comment.

qbarrand: @mhrivnak thanks a lot for reviewing. These are great points and we should indeed take the OpenShift defaults for those values. Do you think that they should be command-line flags as well, like `LeaseDuration` in this PR?  
Thanks!

/hold

cmd/cli/cli.go
```diff
@@ -0,0 +1,38 @@
...
+
+	fs.StringVar(&cl.MetricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
+	fs.BoolVar(&cl.EnableLeaderElection, "enable-leader-election", false,
+		"Enable leader election for controller manager. "+
+			"Enabling this will ensure there is only one active controller manager.")
+	fs.DurationVar(
+		&cl.LeaderElectionLeaseDuration,
+		"leader-election-lease-duration",
+		openShiftDefaultLeaderElection.LeaseDuration.Duration,
+		"The leader election lease duration.")
```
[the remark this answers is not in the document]
qbarrand: The `flag` package already generates a rather useful usage message that includes the default value:
```
$> ./manager --help
Usage of /tmp/bin/manager:
  -enable-leader-election
    	Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.
  -leader-election-lease-duration duration
    	The leader election lease duration. (default 2m17s)
  -metrics-addr string
    	The address the metric endpoint binds to. (default ":8080")
```
I think this is reasonable - WDYT?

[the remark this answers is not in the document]
mhrivnak: Ah nice, I didn't realize it did that automatically. LGTM.

main.go
mhrivnak: It seems like that could get complex quickly for the user to even understand, but probably exposing them all is the right thing to do. I wonder what other openshift operators do; it would be useful to have a look at a couple of others to see what's common, and if there's any consensus about the config interface for these values.

mhrivnak: Browsing results from [a search of openshift projects](https://github.com/search?q=org%3Aopenshift+RenewDeadline&type=code), it looks like the standard practice is to hard-code these durations, and only use a CLI option to enable or disable leader election. I think that makes sense, as its unusual to have a reason at runtime to modify these values. And we could always add CLI options later if/when a use case arises.

WDYT?

qbarrand: @mhrivnak Agreed, I came to the same conclusion. I have started reworking this PR in the upstream repo to instead bring a function that sets OpenShift leader election related defaults to the controller options; that way, we can only call that function when running in OpenShift clusters.
However, I would leave in this PR the refactoring of the CLI flag parsing (removing the lease duration), as it was already merged upstream and brings unit tests. WDYT?

mhrivnak: Makes sense to me.

qbarrand:
@mhrivnak @liqcui the `ApplyOpenShiftOptions` function now applies all `library-go` leader-election defaults to the manager options before starting it. The `leader-election-lease-duration` command-line flag was removed to align with what most other operators offer.  
PTAL.

mhrivnak (approved):

PR #74, opened 2021-11-12
Update README URL for operator description

bthurber:


bthurber:
Need to update config/manifests/bases/openshift-special-resource-operator.clusterserviceversion.yaml as well.  Will submit a new PR.

PR #81, opened 2021-11-24
Move to operator-sdk v1.13.1

qbarrand:
Co-authored-by: David Gray <dagray@redhat.com>

Upstream PR: https://github.com/openshift-psap/special-resource-operator/pull/116


ybettan:
```couldn't get a list of nodes by role (worker): nodes is forbidden: User "system:serviceaccount:ci-op-fwxhqx81:default" cannot list resource "nodes" in API group "" at the cluster scope```

Why do we have k8s-api errors in the unit-tests?
@qbarrand

qbarrand:
@ybettan I think I just found the reason.  
Before this PR, the `test` target was skipped because it was not phony and a `test` directory existed; same problem as in the [upstream PR](https://github.com/openshift-psap/special-resource-operator/pull/121#issuecomment-953958337).  
That same PR added `e2e` build tags to the integration tests, so that they are skipped when running `make test`. The PR was not backported into downstream, so the tags are not there, so the e2e tests are ran, so we see problems related to the Kubernetes API 😃
/hold until I backport https://github.com/openshift-psap/special-resource-operator/pull/121 as well.

PR #88, opened 2021-12-22
Bug 2034766: Special Resource Operator(SRO) - no cert-manager pod created in dual stack environment

pacevedom:
Failure to retrieve the pull secret should not return an early error, as it is only found after creating the namespace and this operation has not been done yet.


qbarrand:
I don't get why CI passes today without that change.
Also, do we want to run this upstream with the e2e CI before we merge it downstream?

pacevedom:
> I don't get why CI passes today without that change. Also, do we want to run this upstream with the e2e CI before we merge it downstream?

You mean the ART issue? That is not run here, thats why it passes. The upstream does not have this issue, it is already using the `warn` package.

qbarrand:
> The upstream does not have this issue, it is already using the warn package.

That explains. Thanks for clarifying.
/lgtm

PR #90, opened 2022-01-05
Bug 2034766: Fix hash annotation equal when updating recipe resources

pacevedom:


pacevedom:
cc @qbarrand @yevgeny-shnaidman @pmtk

PR #95, opened 2022-01-28
Bring recent changes from upstream

qbarrand:
This PR brings a number of commits from upstream, with conflicts resolved where required.


qbarrand:
Hold because of code freeze.
/hold

bthurber:
@qbarrand This will need work before being able to merge.  SInce we are at code freeze today for 4.10 we need to discuss options for targeting a z-stream release.

qbarrand:
@pacevedom PTAL

qbarrand:
@bthurber do we want to include this PR into 4.10 at all? That's quite a lot of changes for a z-stream release, even though they are mostly about testing and refactoring.

PR #96, opened 2022-01-28
Fix poll pkg bug for buildconfig resources

pacevedom:
When there is more than one `buildconfig` resource in the same namespace, poll package queries and waits for all builds, ignoring the build owner, and in turn, possibly waiting on not owned resources. Also fixes waiting for not yet created build pods.

### DIFF
--- a/pkg/poll/poll.go
+++ b/pkg/poll/poll.go
@@ -411,14 +411,27 @@ func ForBuild(obj *unstructured.Unstructured) error {
 		return errors.Wrap(err, "Could not get BuildList")
 	}
 
-	for _, build := range builds.Items {
-		callback := makeStatusCallback(&build, "Complete", "status", "phase")
-		if err := ForResourceFullAvailability(&build, callback); err != nil {
+	var build *unstructured.Unstructured
+	for _, b := range builds.Items {
+		slice, _, err := unstructured.NestedSlice(b.Object, "metadata", "ownerReferences")
+		if err != nil {
 			return err
 		}
+		for _, element := range slice {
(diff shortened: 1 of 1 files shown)


pkg/poll/poll.go
```diff
@@ -411,14 +411,27 @@ func ForBuild(obj *unstructured.Unstructured) error {
...
-		callback := makeStatusCallback(&build, "Complete", "status", "phase")
-		if err := ForResourceFullAvailability(&build, callback); err != nil {
+	var build *unstructured.Unstructured
+	for _, b := range builds.Items {
+		slice, _, err := unstructured.NestedSlice(b.Object, "metadata", "ownerReferences")
+		if err != nil {
 			return err
 		}
+		for _, element := range slice {
+			if name, ok := element.(map[string]interface{})["name"]; ok && name == obj.GetName() {
```
qbarrand, line 421: If the key does not exist in the map, the value will be the empty string. I'd say `ok` is redundant here.
```suggestion
			if element.(map[string]interface{})["name"]== obj.GetName() {
```

qbarrand (commented):
Besides one minor suggestion, lgtm 🙂

qbarrand:
Would be great if you could pull the tests from https://github.com/openshift-psap/special-resource-operator/pull/241 as well.

pacevedom:
> Would be great if you could pull the tests from [openshift-psap#241](https://github.com/openshift-psap/special-resource-operator/pull/241) as well.

@qbarrand Would rather do that in a separate commit, the original pkg is quite different after the tests were added.

PR #97, opened 2022-01-28
Bug 2030840: set an initial ClusterOperator status

qbarrand:
This PR makes the application set an initial `ClusterOperator` status as soon as it starts.
It also uses `defer` to make sure that the status is updated when it exits.

Per the BZ, the `ClusterOperator`'s name is also prefixed with `openshift-`.

This PR requires #95 to be merged. Then the first commit (d6402d28941e49f0f3f209a9b50c98794dbe535b) should be deleted from this PR.

/hold

### DIFF
--- a/main.go
+++ b/main.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"fmt"
 	"os"
 
 	srov1beta1 "github.com/openshift-psap/special-resource-operator/api/v1beta1"
@@ -40,6 +41,7 @@ import (
 	sroscheme "github.com/openshift-psap/special-resource-operator/pkg/scheme"
 	"github.com/openshift-psap/special-resource-operator/pkg/storage"
 	"github.com/openshift-psap/special-resource-operator/pkg/upgrade"
+	"github.com/openshift-psap/special-resource-operator/pkg/utils"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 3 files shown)


wabouhamad (commented):
/lgtm

ybettan:
Don't we need this fix upstream?

Do we have some process for cherry-picking PRs from upstream to downstream?

qbarrand:
Hold as we might remove support for `ClusterOperator` objects entirely.

/hold

wabouhamad (commented):
/lgtm

ybettan:
@qbarrand Can we close that PR now that ClusterOperator was removed?

qbarrand:
@ybettan it was removed upstream only, this PR targets downstream. I don't know what we want to do with it though - I think we want to leave it pending until we have figured out updates. @yevgeny-shnaidman any thoughts?

yevgeny-shnaidman:
> @ybettan it was removed upstream only, this PR targets downstream. I don't know what we want to do with it though - I think we want to leave it pending until we have figured out updates. @yevgeny-shnaidman any thoughts?
I think that we don't need to touch ClusterOperator CR in the downstream for now. It will probably be removed in the downstream also, and in any case, any updates to it are not watch by anyone

ybettan:
> and in any case, any updates to it are not watch by anyone

This is why I am wondering why we are keeping it.

yevgeny-shnaidman:
> > and in any case, any updates to it are not watch by anyone
> 
> This is why I am wondering why we are keeping it.

we are keeping it until we make a final decision, it does not bother anyone right now :)

qbarrand:
Closing this after the corresponding BZ was closed.

PR #100, opened 2022-02-04
Remove NFD runtime dependency

pacevedom:
Remove runtime dependency from SRO to NFD labels in nodes. Now information is taken from the nodes `nodeInfo` struct in `status`.
Now NFD dependency comes from the recipes, not SRO.


pacevedom:
cc @qbarrand @yevgeny-shnaidman

qbarrand:
Hold until the downstream CI is merged. 
/hold

pkg/upgrade/upgrade.go
```diff
@@ -79,37 +79,24 @@ func (ci *clusterInfo) GetClusterInfo(ctx context.Context, nodeList *corev1.Node
...
-		if kernelFullVersion, found = labels[labelKernelVersionFull]; !found {
-			return nil, fmt.Errorf("label %s not found is NFD running? Check node labels", labelKernelVersionFull)
+		kernelFullVersion := node.Status.NodeInfo.KernelVersion
+		if len(kernelFullVersion) == 0 {
+			return info, fmt.Errorf("kernel version not found in node %s", node.Name)
 		}
 
-		if clusterVersion, found = labels[labelOSReleaseVersionID]; !found {
-			return nil, fmt.Errorf("label %s not found is NFD running? Check node labels", labelOSReleaseVersionID)
+		clusterVersion, osVersion, osMajor, err := utils.GetOSInfo(node.Status.NodeInfo.OSImage)
```
yevgeny-shnaidman, line 81: clusterVersion in current code: 4.9
in your new ode:                       49
Currently i don't think it matters, because i cannot find where we are actually using this data... 
Maybe we can remove it from the RuntimeInfo?

pkg/cluster/cluster.go
```diff
@@ -136,38 +136,20 @@ func (c *cluster) OSImageURL(ctx context.Context) (string, error) {
...
-	// On OCP >4.7, we can use the NFD label  feature.node.kubernetes.io/system-os_release.RHEL_VERSION label.
-	if rhelVersion, found := labels[os+".RHEL_VERSION"]; found && len(rhelVersion) == 3 {
-		rhelMaj := rhelVersion[0:1]
-		rhelMin := rhelVersion[2:]
-		return "rhel" + rhelMaj, "rhel" + rhelVersion, rhelMaj + "." + rhelMin, nil
-	}
-
-	// On vanilla k8s and older NFD versions, we need RenderOperatingSystem
-	return utils.RenderOperatingSystem(nodeOSrel, nodeOSmaj, nodeOSmin)
+	return "rhel" + nodeOS, "rhel" + nodeOSMajor, nodeOS, nil
```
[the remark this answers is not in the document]
pacevedom: Good catch.

pkg/upgrade/upgrade.go
pacevedom: Some of the charts are using these values, so I guess we can not remove them altogether now that they have been around for some time.

pacevedom:
Just realized a bug, dont remove the hold yet

yevgeny-shnaidman:
> Just realized a bug, dont remove the hold yet

what is the bug?

pkg/kernel/kernel.go
```diff
@@ -129,22 +130,15 @@ func (k *kernelData) IsObjectAffine(obj client.Object) bool {
...
 	// one could easily add driver-kernel-versions for each node.
 	for _, node := range nodeList.Items {
-		labels := node.GetLabels()
-
-		// We only need to check for the key, the value
-		// is available if the key is there
-		short := "feature.node.kubernetes.io/kernel-version.full"
-		if kernelFullVersion, found = labels[short]; !found {
-			return "", errors.New("Label " + short + " not found is NFD running? Check node labels")
+		kernelFullVersion = node.Status.NodeInfo.KernelVersion
```
qbarrand, line 137: Who is populating this `NodeInfo` field? Any idea why we weren't using it before?

pkg/upgrade/upgrade.go
```diff
@@ -79,37 +69,24 @@ func (ci *clusterInfo) GetClusterInfo(ctx context.Context, nodeList *corev1.Node
...
-			nodeOSmin := labels[labelOSReleaseVersionIDMinor]
-			info[kernelFullVersion] = NodeVersion{OSVersion: nodeOSmaj + "." + nodeOSmin, OSMajor: nodeOSrel + nodeOSmaj, OSMajorMinor: nodeOSrel + nodeOSmaj + "." + nodeOSmin, ClusterVersion: clusterVersion}
-		} else {
-			rhelMaj := rhelVersion[0:1]
-			info[kernelFullVersion] = NodeVersion{OSVersion: rhelVersion, OSMajor: "rhel" + rhelMaj, OSMajorMinor: "rhel" + rhelVersion, ClusterVersion: clusterVersion}
+		info[kernelFullVersion] = NodeVersion{
+			OSVersion:      osVersion,
+			ClusterVersion: clusterVersion,
+			OSMajor:        fmt.Sprintf("rhel%s", osMajor),
+			OSMajorMinor:   fmt.Sprintf("rhel%s", osVersion),
```
qbarrand, line 89:
```suggestion
			OSMajor:        "rhel" + osMajor,
			OSMajorMinor:   "rhel" + osVersion,
```

pkg/cluster/cluster.go
```diff
@@ -136,38 +136,20 @@ func (c *cluster) OSImageURL(ctx context.Context) (string, error) {
...
-		nodeOSmin = labels[os+".VERSION_ID.minor"]
-
-		if len(nodeOSrel) == 0 || len(nodeOSmaj) == 0 {
-			return "", "", "", fmt.Errorf("Cannot extract %s.*, is NFD running? Check node labels", os)
+		_, osVersion, osMajor, err := utils.GetOSInfo(node.Status.NodeInfo.OSImage)
+		if err != nil {
+			return "", "", "", fmt.Errorf("Unable to get node %s OS image info: %w", node.Name, err)
 		}
+		nodeOS = osVersion
+		nodeOSMajor = osMajor
```
qbarrand, line 150:
```suggestion
		var err error // this could go out of the loop

		_, nodeOS, nodeOSMajor, err = utils.GetOSInfo(node.Status.NodeInfo.OSImage)
		if err != nil {
			return "", "", "", fmt.Errorf("unable to get node %s OS image info: %w", node.Name, err)
		}
```

pkg/utils/osversion.go
```diff
@@ -1,33 +1,26 @@
...
+	versionRegex = regexp.MustCompile(`(?P<ClusterVersion>\d{2})\.(?P<OSVersion>\d{2})\.\d{12}-\d`)
+)
+
+// GetOSInfo takes a pretty format string for the OS version (e.g. Red Hat Enterprise Linux CoreOS 49.84.202201102104-0 (Ootpa))
+// and returns the cluster version, OS version in \d\.\d format, plus major version for the OS.
+func GetOSInfo(osImagePretty string) (string, string, string, error) {
+	clusterVersion, osVersion, osMajor := "", "", ""
+	matches := versionRegex.FindStringSubmatch(osImagePretty)
+	if len(matches) != 3 {
+		return clusterVersion, osVersion, osMajor, errors.New("Unable to find a match " + osImagePretty)
```
qbarrand, line 20:
```suggestion
		return clusterVersion, osVersion, osMajor, errors.New("unable to find a match " + osImagePretty)
```

pkg/cluster/cluster_test.go
```diff
@@ -295,46 +295,53 @@ var _ = Describe("cluster_OSImageURL", func() {
 })
 
 var _ = Describe("cluster_OperatingSystem", func() {
-
-	It("should return an error when feature.node.kubernetes.io/system-os_release.ID is empty", func() {
-		nodesList := utils.CreateNodesList(1, nil)
-		_, _, _, err := cluster.NewCluster(nil).OperatingSystem(nodesList)
-		Expect(err).To(HaveOccurred())
+	It("should work when node info OSImage", func() {
```
qbarrand, line 298: I think we could use a more descriptive test case name here.

pkg/cluster/cluster_test.go
```diff
@@ -295,46 +295,53 @@ var _ = Describe("cluster_OSImageURL", func() {
...
-			"feature.node.kubernetes.io/system-os_release.RHEL_VERSION":     "123",
+	It("should fail when node has no OSImage", func() {
+		nodesList := corev1.NodeList{
+			Items: []corev1.Node{
+				{
+					Status: corev1.NodeStatus{
+						NodeInfo: corev1.NodeSystemInfo{},
+					},
+				},
+			},
```
qbarrand, line 341:
```suggestion
			Items: make([]corev1.Node, 1),
```

pkg/kernel/kernel_test.go
```diff
@@ -190,3 +191,36 @@ var _ = Describe("PatchVersion", func() {
...
+	})
+	It("should trigger an error if kernel not present", func() {
+		nodeList := corev1.NodeList{
+			Items: []corev1.Node{
+				{
+					Status: corev1.NodeStatus{
+						NodeInfo: corev1.NodeSystemInfo{},
+					},
+				},
+			},
```
qbarrand, line 221:
```suggestion
			Items: make([]corev1.Node, 1),
```

qbarrand (commented):
Few nits

pkg/kernel/kernel.go
pacevedom: Its the Kubelet. No idea why it was not used before.

pacevedom:
> > Just realized a bug, dont remove the hold yet
> 
> what is the bug?

Was not supporting double digit OCP versions in the regex, like 4.10 for example.

ybettan:
/unhold
d/s CI is operational

pacevedom:
PTAL @qbarrand @pmtk @yevgeny-shnaidman @ybettan @enriquebelarte

pkg/upgrade/upgrade.go
```diff
@@ -84,37 +74,24 @@ func (ci *clusterInfo) GetClusterInfo(ctx context.Context, nodeList *corev1.Node
...
-		var clusterVersion string
-
-		labels := node.GetLabels()
-		// We only need to check for the key, the value
-		// is available if the key is there
-		if kernelFullVersion, found = labels[labelKernelVersionFull]; !found {
-			return nil, fmt.Errorf("label %s not found is NFD running? Check node labels", labelKernelVersionFull)
+		kernelFullVersion := node.Status.NodeInfo.KernelVersion
+		if len(kernelFullVersion) == 0 {
+			return info, fmt.Errorf("kernel version not found in node %s", node.Name)
```
qbarrand, line 83:
```suggestion
			return nil, fmt.Errorf("kernel version not found in node %s", node.Name)
```

pkg/upgrade/upgrade.go
```diff
@@ -84,37 +74,24 @@ func (ci *clusterInfo) GetClusterInfo(ctx context.Context, nodeList *corev1.Node
...
 		}
 
-		if clusterVersion, found = labels[labelOSReleaseVersionID]; !found {
-			return nil, fmt.Errorf("label %s not found is NFD running? Check node labels", labelOSReleaseVersionID)
+		clusterVersion, osVersion, osMajor, err := utils.GetOSInfo(node.Status.NodeInfo.OSImage)
+		if err != nil {
+			return info, err
```
qbarrand, line 88:
```suggestion
			return nil, err
```

pkg/upgrade/upgrade_test.go
```diff
@@ -83,20 +85,11 @@ var _ = Describe("ClusterInfo", func() {
 	systemMajor := "8"
 	systemMinor := "4"
 	clusterVersion := "4.9"
+	osVersion := "Red Hat Enterprise Linux CoreOS %s.%s.202201102104-0 (Ootpa)"
```
qbarrand, line 88:
```suggestion
	const osVersion = "Red Hat Enterprise Linux CoreOS %s.%s.202201102104-0 (Ootpa)"
```

pkg/upgrade/upgrade_test.go
```diff
@@ -105,11 +98,26 @@ var _ = Describe("ClusterInfo", func() {
 	}
 
 	Context("has all required data (happy flow)", func() {
+
+		nodeInfoRTKernel := corev1.NodeSystemInfo{
+			KernelVersion: kernelRT,
+			OSImage:       fmt.Sprintf(osVersion, strings.ReplaceAll(clusterVersion, ".", ""), fmt.Sprintf("%s%s", systemMajor, systemMinor)),
```
qbarrand, line 104: Could this be a plain string with the expected value? I think it would make the test easier to understand.

pkg/upgrade/upgrade_test.go
```diff
@@ -222,28 +227,30 @@ var _ = Describe("ClusterInfo", func() {
 
 	Context("lacks some required data/is mismatched", func() {
 		badSystemMinor := "0"
-		badKernel := ""
-		badNodeOSLabelsWithRTKernel := map[string]string{
-			labelKernelVersionFull:    kernelRT,
-			labelOSReleaseVersionID:   clusterVersion,
-			labelOSReleaseRHELVersion: fmt.Sprintf("%s.%s", systemMajor, badSystemMinor),
+		badKernel := "someotherkernel"
```
qbarrand, line 230:
```suggestion
		const badKernel = "someotherkernel"
```

pkg/utils/osversion.go
```diff
@@ -1,33 +1,26 @@
...
+	"github.com/pkg/errors"
+)
+
+var (
+	versionRegex = regexp.MustCompile(`(?P<ClusterVersion>\d{2,})\.(?P<OSVersion>\d{2,})\.\d{12}-\d`)
+)
+
+// GetOSInfo takes a pretty format string for the OS version (e.g. Red Hat Enterprise Linux CoreOS 49.84.202201102104-0 (Ootpa))
+// and returns the cluster version, OS version in \d\.\d format, plus major version for the OS.
+func GetOSInfo(osImagePretty string) (string, string, string, error) {
```
qbarrand, line 16: I think `ParseOSInfo` would be a better name for this function.

pkg/kernel/kernel_test.go
```diff
@@ -190,3 +191,30 @@ var _ = Describe("PatchVersion", func() {
 		Entry(nil, "4.18.0-305", "4.18.0-305"),
 	)
 })
+
+var _ = Describe("FullVersion", func() {
+	It("should return the version from the node", func() {
+		kernelVersion := "4.18.0-305.30.1.el8_4.x86_64"
```
qbarrand, line 197:
```suggestion
		const kernelVersion = "4.18.0-305.30.1.el8_4.x86_64"
```

pkg/upgrade/upgrade.go
```diff
@@ -84,37 +74,24 @@ func (ci *clusterInfo) GetClusterInfo(ctx context.Context, nodeList *corev1.Node
 }
 
 func (ci *clusterInfo) nodeVersionInfo(nodeList *corev1.NodeList) (map[string]NodeVersion, error) {
-
-	var found bool
-	var info = make(map[string]NodeVersion)
-
+	info := make(map[string]NodeVersion)
```
qbarrand, line 77:
```suggestion
	info := make(map[string]NodeVersion, len(nodeList.Items))
```

pacevedom: Nodes usually (unless in the middle of an upgrade) have the same kernel, therefore preallocating space might result in a waste. By the same, reallocations should be rare.

pkg/upgrade/upgrade_test.go
pacevedom: Technically it can, but if you change constants (for whatever reason) you need to manually update this too, which implies looking at formats and so on. I dont have a strong opinion on neither of the different approaches.

pkg/upgrade/upgrade.go
qbarrand: Right, that makes sense.

pkg/upgrade/upgrade_test.go
qbarrand: I think it would reduce logic in the tests and make them more assertive moving forward. I would do it.

pacevedom: Mind if I change the approach to have the `osVersion` without dots plus the `systemMajor+systemMinor` defined right below the constants? It is used in 6 different places.

qbarrand: I think that's reasonable.

qbarrand:
/lgtm

🥳

PR #104, opened 2022-02-07
Use DTK ImageStream to get an DTK image URL

pmtk:
Instead of fetching and unpacking release image to get the DTK Image URL, a `driver-toolkit` `ImageStream` is used.

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -10,6 +10,7 @@ import (
 	"github.com/openshift-psap/special-resource-operator/pkg/clients"
 	"github.com/openshift-psap/special-resource-operator/pkg/utils"
 	configv1 "github.com/openshift/api/config/v1"
+	imagev1 "github.com/openshift/api/image/v1"
 	machinev1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 
 	corev1 "k8s.io/api/core/v1"
@@ -24,9 +25,9 @@ import (
 
 type Cluster interface {
 	Version(context.Context) (string, string, error)
-	VersionHistory(context.Context) ([]string, error)
 	OSImageURL(context.Context) (string, error)
 	OperatingSystem(*corev1.NodeList) (string, string, string, error)
+	GetDTKImages(context.Context) ([]string, error)
 }
 
 func NewCluster(clients clients.ClientsInterface) Cluster {
@@ -39,6 +40,27 @@ func NewCluster(clients clients.ClientsInterface) Cluster {
 type cluster struct {
 	log     logr.Logger
 	clients clients.ClientsInterface
+}
+
+func (c *cluster) GetDTKImages(ctx context.Context) ([]string, error) {
+	is := imagev1.ImageStream{}
+
+	err := c.clients.Get(ctx,
+		types.NamespacedName{Namespace: "openshift", Name: "driver-toolkit"},
+		&is)
+	if err != nil {
+		return []string{}, fmt.Errorf("could not obtain openshift/driver-toolkit ImageStream: %w", err)
+	}
+
+	imgs := []string{}
+	for _, tag := range is.Status.Tags {
+		if tag.Tag != "latest" {
+			// DockerImageReference points directly to quay.io
+			imgs = append(imgs, tag.Items[0].DockerImageReference)
+		}
+	}
+
+	return imgs, nil
 }
 
 func (c *cluster) Version(ctx context.Context) (string, string, error) {
@@ -74,34 +96,6 @@ func (c *cluster) Version(ctx context.Context) (string, string, error) {
 	}
 
 	return "", "", errors.New("Undefined Cluster Version")
-}
-
-func (c *cluster) VersionHistory(ctx context.Context) ([]string, error) {
-
-	stat := []string{}
-
-	available, err := c.clusterVersionAvailable()
-	if err != nil {
-		return nil, err
-	}
-	if !available {
-		return stat, nil
-	}
-
-	version, err := c.clients.ClusterVersionGet(ctx, metav1.GetOptions{})
-	if err != nil {
-		return stat, fmt.Errorf("ConfigClient unable to get ClusterVersions: %w", err)
-	}
-
-	stat = append(stat, version.Status.Desired.Image)
-
-	for _, condition := range version.Status.History {
-		if condition.State == "Completed" {
-			stat = append(stat, condition.Image)
-		}
-	}
-
-	return stat, nil
 }
 
 func (c *cluster) OSImageURL(ctx context.Context) (string, error) {
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sort"
 	"strings"
+	"time"
 
 	"github.com/go-logr/logr"
 	"github.com/openshift-psap/special-resource-operator/pkg/clients"
@@ -42,6 +44,7 @@ type cluster struct {
 	clients clients.ClientsInterface
 }
 
+// GetDTKImages returns URLs to DTK images obtained from cluster's DTK ImageStream
 func (c *cluster) GetDTKImages(ctx context.Context) ([]string, error) {
 	is := imagev1.ImageStream{}
 
@@ -52,15 +55,30 @@ func (c *cluster) GetDTKImages(ctx context.Context) ([]string, error) {
 		return []string{}, fmt.Errorf("could not obtain openshift/driver-toolkit ImageStream: %w", err)
 	}
 
-	imgs := []string{}
+	type tagRef struct {
+		ref     string
+		created time.Time
+	}
+
+	trs := []tagRef{}
 	for _, tag := range is.Status.Tags {
-		if tag.Tag != "latest" {
-			// DockerImageReference points directly to quay.io
-			imgs = append(imgs, tag.Items[0].DockerImageReference)
+		if tag.Tag == "latest" {
+			for _, t := range tag.Items {
+				trs = append(trs, tagRef{ref: t.DockerImageReference, created: t.Created.Time})
+			}
 		}
 	}
 
-	return imgs, nil
+	sort.Slice(trs, func(i, j int) bool {
+		return trs[i].created.After(trs[j].created)
+	})
+
+	refs := []string{}
+	for _, tr := range trs {
+		refs = append(refs, tr.ref)
+	}
+
+	return refs, nil
 }
 
 func (c *cluster) Version(ctx context.Context) (string, string, error) {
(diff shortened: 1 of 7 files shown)


pmtk:
Things to note:
New Tag for DTK ImageStream appears about 20 minutes after start of cluster upgrade. Between "new ImageStreamTag" and "upgrade completion" (all nodes (incl. masters) are running newer kernel) is time when SRO will report that "nodes are not running DTK kernel". Right now, this might also occur: ClusterVersion changes at the very end, so until that moment, older DTK will be fetched and compare against newer kernels (for nodes that already got upgraded).
To change this behavior some functions (e.g. `clusterInfo.driverToolkitVersion()`) would have to be reworked to use all items rather than just first.

pkg/cluster/cluster.go
```diff
@@ -41,6 +44,43 @@ type cluster struct {
...
 
+// GetDTKImages returns URLs to DTK images obtained from cluster's DTK ImageStream
+func (c *cluster) GetDTKImages(ctx context.Context) ([]string, error) {
+	is := imagev1.ImageStream{}
+
+	err := c.clients.Get(ctx,
+		types.NamespacedName{Namespace: "openshift", Name: "driver-toolkit"},
+		&is)
+	if err != nil {
+		return []string{}, fmt.Errorf("could not obtain openshift/driver-toolkit ImageStream: %w", err)
```
qbarrand, line 55:
```suggestion
		return nil, fmt.Errorf("could not obtain openshift/driver-toolkit ImageStream: %w", err)
```

pkg/upgrade/upgrade.go
```diff
@@ -179,47 +174,30 @@ func (ci *clusterInfo) updateInfo(info map[string]NodeVersion, dtk registry.Driv
...
-func (ci *clusterInfo) driverToolkitVersion(ctx context.Context, entries []string, info map[string]NodeVersion) (map[string]NodeVersion, error) {
-	if len(entries) == 0 {
+func (ci *clusterInfo) driverToolkitVersion(ctx context.Context, dtkImages []string, info map[string]NodeVersion) (map[string]NodeVersion, error) {
+	if len(dtkImages) == 0 {
 		return info, nil
 	}
 
 	var dtk registry.DriverToolkitEntry
-	var imageURL string
+	imageURL := dtkImages[0]
```
qbarrand, line 183: Do I understand correctly that we're always using the first reference in the slice, i.e. the most recent one?
Should we make `GetDTKImages` return only one reference then?

pkg/upgrade/upgrade.go
```diff
@@ -26,6 +25,8 @@ const (
 	labelOSReleaseVersionIDMinor = "feature.node.kubernetes.io/system-os_release.VERSION_ID.minor"
 )
 
+type dtkImageURL = string
```
qbarrand, line 28: Do we need this alias? Shouldn't variable names when using the `cache` map be enough to understand the nature of the key?

pkg/cluster/cluster_test.go
```diff
@@ -338,3 +262,58 @@ var _ = Describe("cluster_OperatingSystem", func() {
...
+		urls, err := cluster.NewCluster(mockKubeClients).GetDTKImages(context.TODO())
+		Expect(err).To(HaveOccurred())
+		Expect(urls).To(HaveLen(0))
+	})
+
+	It("should return sorted slice of URLs", func() {
+		remoteRegistryURL := "reg.io/release/repo"
+		img1 := "sha256:1"
+		img2 := "sha256:2"
+		img3 := "sha256:3"
```
qbarrand, line 281:
```suggestion
		const (
			remoteRegistryURL = "reg.io/release/repo"
			img1              = "sha256:1"
			img2              = "sha256:2"
			img3              = "sha256:3"
		)
```

pkg/upgrade/upgrade.go
```diff
@@ -69,18 +65,17 @@ func (ci *clusterInfo) GetClusterInfo(ctx context.Context, nodeList *corev1.Node
 		return nil, fmt.Errorf("failed to get upgrade info: %w", err)
 	}
 
-	history, err := ci.cluster.VersionHistory(ctx)
+	dtkImages, err := ci.cluster.GetDTKImages(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("could not get version history: %w", err)
```
qbarrand, line 70:
```suggestion
		return nil, fmt.Errorf("could not get DTK images: %w", err)
```

pkg/upgrade/upgrade.go
```diff
@@ -179,47 +174,30 @@ func (ci *clusterInfo) updateInfo(info map[string]NodeVersion, dtk registry.Driv
...
-		if err != nil {
-			return nil, fmt.Errorf("could not extract version from payload: %w", err)
-		}
-		if imageURL == "" {
-			utils.WarnOnError(errors.New("No DTK image found, DTK cannot be used in a Build"))
-			return info, nil
-		}
-		if layer, err = ci.registry.LastLayer(ctx, imageURL); layer == nil {
+		layer, err := ci.registry.LastLayer(ctx, imageURL)
+		if layer == nil {
```
qbarrand, line 190: We are not checking `err` here. Not sure how I missed this in the previous PR.

pkg/upgrade/upgrade.go
```diff
@@ -179,47 +174,30 @@ func (ci *clusterInfo) updateInfo(info map[string]NodeVersion, dtk registry.Driv
...
 			return nil, fmt.Errorf("cannot extract last layer for DTK from %s: %w", imageURL, err)
 		}
 
 		dtk, err = ci.registry.ExtractToolkitRelease(layer)
```
qbarrand, line 194: It would be very convenient if that method returned a `*DriverToolkitEntry`. WDYT? Should it be included in this PR?

pkg/cluster/cluster.go
```diff
@@ -41,6 +44,43 @@ type cluster struct {
...
+				trs = append(trs, tagRef{ref: t.DockerImageReference, created: t.Created.Time})
+			}
+		}
+	}
+
+	sort.Slice(trs, func(i, j int) bool {
+		return trs[i].created.After(trs[j].created)
+	})
+
+	refs := []string{}
```
qbarrand, line 76:
```suggestion
	refs := make([]string, 0, len(trs))
```

pkg/cluster/cluster_test.go
```diff
@@ -338,3 +262,58 @@ var _ = Describe("cluster_OperatingSystem", func() {
...
+
+var _ = Describe("cluster_GetDTKImages", func() {
+	It("should return an error in case of GET failure", func() {
+		mockKubeClients.EXPECT().
+			Get(gomock.Any(), types.NamespacedName{Namespace: "openshift", Name: "driver-toolkit"}, gomock.Any()).
+			Return(randomError)
+
+		urls, err := cluster.NewCluster(mockKubeClients).GetDTKImages(context.TODO())
+		Expect(err).To(HaveOccurred())
+		Expect(urls).To(HaveLen(0))
```
qbarrand, line 274: I think we can drop that one. If the function returns an error, the calling code should not use other return values.

pmtk: Dropped :)

pkg/upgrade/upgrade.go
pmtk: Removed

pkg/upgrade/upgrade.go
pmtk: Yes, before Pablo's change it was a `for` loop that would in happy flow do only 1 iteration (it'd only advance to second entry if `layer == nil`).
I think `GetDTKImages` returning a slice might be helpful during upgrade scenarios (to fetch DTKs for different cluster versions).

pkg/upgrade/upgrade.go
pmtk: Nice catch, added error checking

pkg/upgrade/upgrade.go
pmtk: Sure, I can include it

pkg/upgrade/upgrade.go
yevgeny-shnaidman: I think that we might have a small chance for ace condition here in case of upgrade, but it is not related to @pmtk CR. 
in case a CR is created just as an upgrade begins, our code will fill the values only for the new cluster version , but not for the old one. So a daemon set will be created for the new nodes, but not the old ones. In bug enough cluster it might be an issue.  @pmtk maybe this scenario is worth investigating?

pkg/upgrade/upgrade.go
```diff
@@ -179,53 +172,39 @@ func (ci *clusterInfo) updateInfo(info map[string]NodeVersion, dtk registry.Driv
...
+	imageURL := dtkImages[0]
 
-	entry := entries[0]
-
-	if cached := ci.cache[entry]; cached != nil {
-		dtk = cached.DTK
-		imageURL = cached.ImageURL
-		ci.log.Info("History from cache", "entry", entry, "dtk", dtk, "imageURL", imageURL)
+	if cached := ci.cache[imageURL]; cached != nil {
+		dtk = cached
```
qbarrand, line 184: So now I think we can do this:
```suggestion
	if dtk = ci.cache[imageURL]; dtk != nil {
```

pkg/upgrade/upgrade.go
pmtk: _I have edited my comment from the day of PR creation because I was mistaken about when the `clusterversion` `.status.desired.image` is updated._

I checked the timeline of the cluster upgrade:
```
3 master, 3 worker, 4.9.9 -> 4.9.15

0 min - cluster upgrade starts, clusterversion .status.desired.image update
24 min - dtk imagestream updated
43 min - 1st worker & master upgrade START
46 min - 1st worker end
46 min - 2nd worker start
47 min - 1st master upgrade end
47 min - 2nd master upgrade start
50 min - 2nd worker end
52 min - 2nd master upgrade end
50 min - 3rd worker start
52 min - 3rd master upgrade start
52 min - 3rd worker end
57 min - 3rd master upgrade end
58 min - cluster upgrade finish
```

Currently, using `.status.desired.image` SRO will detect a mismatch between DTK and node's kernels from the beginning of the upgrade until the specific node is upgraded. 
Using imagestream this gap is halved, but still exists.

Since behavior of taking just first entry from the cluster history is present all this time (just dressed up as a loop with return at the end), I think this is definitely worth investigating but I think it's strongly connected to the cluster upgrade scenarios.

pkg/upgrade/upgrade.go
pmtk: haha, absolutely :D

pkg/upgrade/upgrade.go
yevgeny-shnaidman: agree. lets just open a Jira for this issue

pacevedom: > I think that we might have a small chance for ace condition here in case of upgrade, but it is not related to @pmtk CR. in case a CR is created just as an upgrade begins, our code will fill the values only for the new cluster version , but not for the old one. So a daemon set will be created for the new nodes, but not the old ones. In bug enough cluster it might be an issue. @pmtk maybe this scenario is worth investigating?

Problem with this is that having a heterogeneous situation like the one you describe depends on the way the recipe was written. You would need 2 daemonsets and 2 buildconfigs (potentially) depending on the node they need to land on, or else only one will be created and only those matching nodes will be schedulable.

PR #105, opened 2022-02-07
MGMT-9193: Create metric for upgrade alert

yevgeny-shnaidman:
Define and create a metric that will be later used by the "pre-flight"
upgrade code to trigger alert in case SRO encounters a problem with
driver-container for a specific CR

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -11,6 +11,7 @@ const (
 	completedStatesQuery         = "sro_states_completed_info"
 	completedKindQuery           = "sro_kind_completed_info"
 	usedNodesQuery               = "sro_used_nodes"
+	upgradeAlertQuery            = "sro_upgrade_alert"
 )
 
 var (
@@ -42,6 +43,14 @@ var (
 		},
 		[]string{"cr", "kind", "name", "namespace", "nodes"},
 	)
+
+	upgradeAlert = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: upgradeAlertQuery,
+			Help: "For a SRO CR, 1 if during upgrade there is a problem with CR, 0 otherwise",
+		},
+		[]string{"cr"},
+	)
 )
 
 func init() {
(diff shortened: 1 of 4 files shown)


pkg/metrics/metrics.go
```diff
@@ -42,6 +43,15 @@ var (
 		},
 		[]string{"cr", "kind", "name", "namespace", "nodes"},
 	)
+
+	upgradeAlert = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: upgradeAlertQuery,
+			Help: "For a SRO CR, 1 if during upgrade there is a problem with CR, 0 otherwise",
```
ybettan, line 50: Can't we just a have a single counter for all CRs that can be differentiated by labels?
What is the point of having such metric with `0` value? will be the same as not having a metric at all no?!

qbarrand: I guess the idea is that you can then include the CR name in alerts?

qbarrand (commented):
How about we defer this change until the metric is actually used? Otherwise there's a risk that this code will never be called.

pkg/metrics/metrics.go
yevgeny-shnaidman: we want the alerts to be meaningful, so it needs to get data from the metric. currently it is the CR name, but later we can expand it to other label, if need arises

yevgeny-shnaidman:
> How about we defer this change until the metric is actually used? Otherwise there's a risk that this code will never be called.

yes.

yevgeny-shnaidman:
> How about we defer this change until the metric is actually used? Otherwise there's a risk that this code will never be called.

yes

qbarrand:
> yes

/hold

yevgeny-shnaidman:
> How about we defer this change until the metric is actually used? Otherwise there's a risk that this code will never be called.

actually this a preparation for the next PR, that will use this code to signal that there is a problem in the upgrade. I don't want it to be a part of the same PR

pkg/metrics/metrics.go
ybettan: But we need to alert only in case the value is `1` no?! If that the case, we don't need to have a metric when the value is `0`.

PR #110, opened 2022-02-10
MGMT-9251: Fix state machine loop regarding stateless yamls

yevgeny-shnaidman:
This PR removes processing of the stateless yaml from the statefull
yamls state machine loop


controllers/resources.go
```diff
@@ -159,7 +159,8 @@ func ReconcileChartStates(ctx context.Context, r *SpecialResourceReconciler) err
 		state.GenerateName(stateYAML, r.specialresource.Name)
 
 		step := nostate
-		step.Templates = append(nostate.Templates, stateYAML)
+		step.Templates = []*chart.File{}
+		step.Templates = append(step.Templates, stateYAML)
```
qbarrand, line 163:
```suggestion
		step.Templates = []*chart.File{stateYAML}
```

yevgeny-shnaidman: Fixed

pacevedom (approved):
/lgtm

pkg/assets/assets.go
```diff
@@ -7,12 +7,15 @@ import (
...
 	"path/filepath"
 	"regexp"
+	"strings"
 
 	"github.com/go-logr/logr"
 	"github.com/openshift-psap/special-resource-operator/pkg/utils"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
 
+const namedTemplatePrefix = "_"
```
qbarrand, line 17: Do we need that constant? It's used only once.

controllers/resources.go
```diff
@@ -256,8 +257,10 @@ func ReconcileChartStates(ctx context.Context, r *SpecialResourceReconciler) err
 		}
 	}
 
-	// We're done with states now execute the part of the chart without
+	// We're done with states, now execute the part of the chart without
 	// states we need to reconcile the nostate Chart
+	nostate := basicChart
```
qbarrand, line 262: Could we create a new chart here, or do we need to keep some of `basicChart`'s values?

controllers/resources.go
```diff
@@ -158,8 +162,8 @@ func ReconcileChartStates(ctx context.Context, r *SpecialResourceReconciler) err
 		// affinity, anti-affinity
 		state.GenerateName(stateYAML, r.specialresource.Name)
 
-		step := nostate
-		step.Templates = append(nostate.Templates, stateYAML)
+		step := basicChart
```
qbarrand, line 165: Could we create a new chart here, or do we need to keep some of basicChart's values?

qbarrand (commented):
LGTM besides these few nits.

controllers/resources.go
yevgeny-shnaidman: i prefer to use an existent one, so i we won't need to initial to worry about initialization of some of the fields of the chart. That way we can only take care of the templates.

controllers/resources.go
yevgeny-shnaidman: i prefer to use an existent one, so i we won't need to initial to worry about initialization of some of the fields of the chart. That way we can only take care of the templates.

controllers/resources.go
```diff
@@ -125,21 +125,25 @@ func createImagePullerRoleBinding(ctx context.Context, r *SpecialResourceReconci
...
-	nostate.Templates = []*chart.File{}
-
+	basicChart := r.chart
+	basicChart.Templates = []*chart.File{}
 	stateYAMLS := []*chart.File{}
+	statelessYAMLS := []*chart.File{}
 
-	// First get all non-state related files from the templates
-	// and save the states in a temporary slice for single execution
+	// diffirentiate between statefull YAMLs,
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

controllers/resources.go
ybettan: Can we please make a deep copy then? some of the fields of the underlaying objects are pointers, we will eventually have 2 different objects that shares some inner fields (but not all of them) - I don't think this is a desirable state.

PR #111, opened 2022-02-15
Use in-memory pull-secret instead of writing to disk

pmtk:
Using pull secret in-memory makes running the operator easier in out-of-cluster scenario (using `go run .` against `$KUBECONFIG`)

### DIFF
--- a/pkg/registry/registry.go
+++ b/pkg/registry/registry.go
@@ -7,10 +7,11 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"os"
+	"net/url"
 	"strings"
 
 	"github.com/go-logr/logr"
+	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/crane"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/openshift-psap/special-resource-operator/pkg/clients"
@@ -23,10 +24,9 @@ import (
 )
 
 const (
-	pullSecretNamespace  = "openshift-config"
-	pullSecretName       = "pull-secret"
-	pullSecretFileName   = ".dockerconfigjson"
-	dockerConfigFilePath = "/home/nonroot/.docker/config.json"
+	pullSecretNamespace = "openshift-config"
+	pullSecretName      = "pull-secret"
+	pullSecretFileName  = ".dockerconfigjson"
 )
 
 var (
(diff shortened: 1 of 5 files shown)


pkg/registry/registry.go
```diff
@@ -60,32 +60,67 @@ type registry struct {
 	log        logr.Logger
 }
 
-func (r *registry) writeImageRegistryCredentials(ctx context.Context) error {
+type dockerAuth struct {
+	Auth  string
+	Email string
+}
+
+func (r *registry) registryFromImageURL(image string) (string, error) {
```
qbarrand, line 68: Could we have a unit test for this function?

pkg/registry/registry.go
```diff
@@ -60,32 +60,67 @@ type registry struct {
...
+	}{}
+
+	err = json.Unmarshal(pullSecretData, &auths)
 	if err != nil {
-		return errors.Wrap(err, "Can not write pull secret file")
+		return dockerAuth{}, errors.New("Failed to unmarshal auths")
+	}
+
+	if auth, ok := auths.Auths[registry]; !ok {
+		return dockerAuth{}, errors.New("Cluster PullSecret does not contain auth for the registry " + registry)
```
qbarrand, line 110:
```suggestion
		return dockerAuth{}, fmt.Errorf("cluster PullSecret does not contain auth for registry %s", registry)
```

pkg/registry/registry.go
```diff
@@ -60,32 +60,67 @@ type registry struct {
...
 	s, err := r.kubeClient.GetSecret(ctx, pullSecretNamespace, pullSecretName, metav1.GetOptions{})
 	if err != nil {
-		return errors.Wrap(err, "Can not retrieve pull secrets")
+		return dockerAuth{}, errors.Wrap(err, "Can not retrieve pull secrets")
 	}
 
 	pullSecretData, ok := s.Data[pullSecretFileName]
 	if !ok {
-		return errors.New("Can not find data content in the secret")
+		return dockerAuth{}, errors.New("Can not find data content in the secret")
```
qbarrand, line 97:
```suggestion
		return dockerAuth{}, errors.New("could not find data content in the secret")
```

pacevedom (commented):
I am missing the changes to drop the volume in the manifests.

pkg/registry/registry.go
pmtk: Yes, added :)

pkg/registry/registry.go
```diff
@@ -60,32 +60,67 @@ type registry struct {
...
+	return u.Host, nil
+}
+
+func (r *registry) getImageRegistryCredentials(ctx context.Context, registry string) (dockerAuth, error) {
 	_, err := r.kubeClient.GetNamespace(ctx, pullSecretNamespace, metav1.GetOptions{})
 	if err != nil {
 		r.log.Info("Can not find namespace for pull secrets, assuming vanilla k8s")
-		return nil
+		return dockerAuth{}, nil
 	}
```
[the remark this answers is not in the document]
pmtk: Removed

PR #114, opened 2022-02-22
Bug 2054787: SRO controller crashes when pull-secret holds bad permissions

pacevedom:


pacevedom:
Had to rebase, some commit overtook this one and it broke everything.

PR #115, opened 2022-02-24
Change location of Helm's cache

pmtk:


qbarrand:
/hold
Reviewing this changeset in https://github.com/openshift-psap/special-resource-operator/pull/265.

pkg/helmer/helmer.go
```diff
@@ -5,6 +5,8 @@ import (
 	"context"
 	"encoding/json"
```
qbarrand, line 6:
```suggestion
	"encoding/json"
	"errors"
```

PR #117, opened 2022-02-25
MGMT-9392: Move controller runtime code to a dedicated package

yevgeny-shnaidman:
This PR moves all the runtime related code into a dedicated package
and removes RunInfo global variables ( it is moved under the
SpecialResourceReconciler structure)


qbarrand:
@yevgeny-shnaidman shouldn't this be implemented upstream first?

yevgeny-shnaidman:
> @yevgeny-shnaidman shouldn't this be implemented upstream first?

I don't see mcuh deference between pushing it to downstream and then to upstream and visa-versa. I am pushing it to downstream since i need this PR in order to proceed with upgrade

pkg/runtime/runtime.go
```diff
@@ -0,0 +1,232 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to get cluster version: %w", err)
+	}
+
+	runInfo.ClusterUpgradeInfo, err = rt.clusterInfoAPI.GetClusterInfo(ctx, nodeList)
+	if err != nil {
+		return fmt.Errorf("failed to get upgrade info: %w", err)
+	}
+
+	fmt.Printf("\nYEV before etryGetPushSecretNam platform is %s\n", runInfo.Platform)
```
pmtk, line 164: Debug string? :)

controllers/upgrade.go
```diff
@@ -19,7 +19,7 @@ func SpecialResourceUpgrade(ctx context.Context, r *SpecialResourceReconciler) (
 		return ctrl.Result{}, fmt.Errorf("failed to get nodes: %w", err)
 	}
 
-	RunInfo.ClusterUpgradeInfo, err = r.ClusterInfo.GetClusterInfo(ctx, nodeList)
+	r.RunInfo.ClusterUpgradeInfo, err = r.ClusterInfo.GetClusterInfo(ctx, nodeList)
```
pmtk, line 22: One thing here: what we're assigning here might be incorrect because it's based on `r.specialresource` from last reconcile (`SpecialResourceUpgrade()` is executed before `SpecialResourcesReconcile()` which fetches SR to handle in current reconcile).
But that's a side note to this PR; possible improvement in future code changes.

pmtk:
Looks good to me, maybe in future we could drop `r.RunInfo` (along with other global vars and god object) and pass "work objects" as arguments

pkg/runtime/runtime.go
yevgeny-shnaidman: yep, sorry. missed it :).Fixed

controllers/upgrade.go
yevgeny-shnaidman: i agree. the best practice would probably be to create a context per reconcile loop and store all the related data in it (RunInfo)

pkg/runtime/runtime.go
```diff
@@ -0,0 +1,229 @@
...
+	rt.log.Info("Runtime Information", "ClusterVersion", runInfo.ClusterVersion)
+	rt.log.Info("Runtime Information", "ClusterVersionMajorMinor", runInfo.ClusterVersionMajorMinor)
+	rt.log.Info("Runtime Information", "ClusterUpgradeInfo", runInfo.ClusterUpgradeInfo)
+	rt.log.Info("Runtime Information", "PushSecretName", runInfo.PushSecretName)
+	rt.log.Info("Runtime Information", "OSImageURL", runInfo.OSImageURL)
+	rt.log.Info("Runtime Information", "Proxy", runInfo.Proxy)
+}
+
+func (rt *runtime) GetRuntimeInformation(ctx context.Context, sr *srov1beta1.SpecialResource, runInfo *RuntimeInformation) error {
+	var err error
```
qbarrand, line 124: I think this declaration can be removed.

pkg/runtime/runtime.go
```diff
@@ -0,0 +1,229 @@
...
+
+	secrets.SetAPIVersion("v1")
+	secrets.SetKind("SecretList")
+
+	rt.log.Info("Getting SecretList in Namespace: " + sr.Spec.Namespace)
+	opts := []client.ListOption{
+		client.InNamespace(sr.Spec.Namespace),
+	}
+	err := rt.kubeClient.List(ctx, secrets, opts...)
+	if err != nil {
```
qbarrand, line 214:
```suggestion
	if err := rt.kubeClient.List(ctx, secrets, client.InNamespace(sr.Spec.Namespace)); err != nil {
```

pkg/runtime/runtime.go
```diff
@@ -0,0 +1,229 @@
...
+	for _, secret := range secrets.Items {
+		secretName := secret.GetName()
+
+		if strings.Contains(secretName, "builder-dockercfg") {
+			rt.log.Info("Found", "Secret", secretName)
+			return secretName, nil
+		}
+	}
+
+	return "", errors.New("Cannot find Secret builder-dockercfg")
```
qbarrand, line 228:
```suggestion
	return "", errors.New("cannot find Secret builder-dockercfg")
```

pkg/runtime/runtime.go
```diff
@@ -0,0 +1,229 @@
...
+		return fmt.Errorf("failed to get kernel patch version: %w", err)
+	}
+
+	// Only want to initialize the platform once.
+	if runInfo.Platform == "" {
+		runInfo.Platform, err = rt.kubeClient.GetPlatform()
+		if err != nil {
+			return fmt.Errorf("failed to determine platform: %v", err)
+		}
+	}
```
qbarrand, line 152: I think we might be able to drop this field entirely. It's only used when getting the push secret, which we can make unconditional since this operator is downstream only.

pkg/runtime/runtime.go
```diff
@@ -0,0 +1,229 @@
...
+func (rt *runtime) getPushSecretName(ctx context.Context, sr *srov1beta1.SpecialResource, platform string) (string, error) {
+	if platform == "K8S" {
+		rt.log.Info("Warning: On vanilla K8s. Skipping search for push-secret")
+		return "", nil
+	}
+
+	secrets := &unstructured.UnstructuredList{}
+
+	secrets.SetAPIVersion("v1")
+	secrets.SetKind("SecretList")
```
qbarrand, line 207: Could we instead use a `v1.SecretList`?

pkg/runtime/runtime.go
```diff
@@ -0,0 +1,229 @@
...
+	rt.log.Info("Runtime Information", "KernelFullVersion", runInfo.KernelFullVersion)
+	rt.log.Info("Runtime Information", "KernelPatchVersion", runInfo.KernelPatchVersion)
+	rt.log.Info("Runtime Information", "DriverToolkitImage", runInfo.DriverToolkitImage)
+	rt.log.Info("Runtime Information", "Platform", runInfo.Platform)
+	rt.log.Info("Runtime Information", "ClusterVersion", runInfo.ClusterVersion)
+	rt.log.Info("Runtime Information", "ClusterVersionMajorMinor", runInfo.ClusterVersionMajorMinor)
+	rt.log.Info("Runtime Information", "ClusterUpgradeInfo", runInfo.ClusterUpgradeInfo)
+	rt.log.Info("Runtime Information", "PushSecretName", runInfo.PushSecretName)
+	rt.log.Info("Runtime Information", "OSImageURL", runInfo.OSImageURL)
+	rt.log.Info("Runtime Information", "Proxy", runInfo.Proxy)
```
qbarrand, line 120: How about this?
```suggestion
	rt.log.Info("Runtime Information",
		"OperatingSystemMajor", runInfo.OperatingSystemMajor,
		"OperatingSystemMajorMinor", runInfo.OperatingSystemMajorMinor,
		"OperatingSystemDecimal", runInfo.OperatingSystemDecimal,
		"KernelFullVersion", runInfo.KernelFullVersion,
		"KernelPatchVersion", runInfo.KernelPatchVersion,
		"DriverToolkitImage", runInfo.DriverToolkitImage,
		"Platform", runInfo.Platform,
		"ClusterVersion", runInfo.ClusterVersion,
		"ClusterVersionMajorMinor", runInfo.ClusterVersionMajorMinor,
		"ClusterUpgradeInfo", runInfo.ClusterUpgradeInfo,
		"PushSecretName", runInfo.PushSecretName,
		"OSImageURL", runInfo.OSImageURL,
		"Proxy", runInfo.Proxy)
```

pkg/runtime/runtime.go
```diff
@@ -0,0 +1,229 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to get Proxy Configuration: %w", err)
+	}
+
+	sr.DeepCopyInto(&runInfo.SpecialResource)
+
+	return nil
+}
+
+func (rt *runtime) retryGetPushSecretName(ctx context.Context, sr *srov1beta1.SpecialResource, platform string) (string, error) {
```
qbarrand, line 182: Why do we even have this logic? I think it might be an optimization for when the namespace was just created and the secret is not there yet, but I find it rather confusing. I would remove this function, return the error if we cannot finf the secret, and ultimately trigger a new reconciliation loop.

pkg/runtime/runtime.go
```diff
@@ -0,0 +1,229 @@
...
+
+	return "", errors.New("Cannot find Secret builder-dockercfg")
+
+}
+
+func (rt *runtime) getPushSecretName(ctx context.Context, sr *srov1beta1.SpecialResource, platform string) (string, error) {
+	if platform == "K8S" {
+		rt.log.Info("Warning: On vanilla K8s. Skipping search for push-secret")
+		return "", nil
+	}
```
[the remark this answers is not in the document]
yevgeny-shnaidman: I thought about that, but kept it  for now. I will remove it

pkg/runtime/runtime.go
yevgeny-shnaidman: The same comment as above: will be easier to port to upstream as a placeholder for other vendors to add their code

pkg/runtime/runtime.go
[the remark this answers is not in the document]
yevgeny-shnaidman: WDYT?

pkg/runtime/runtime.go
yevgeny-shnaidman: Fixed

pkg/runtime/runtime.go
yevgeny-shnaidman: Fixed

pkg/runtime/runtime.go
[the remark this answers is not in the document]
qbarrand: I think upstream will use Shipwright, which is another story. If vendors want to add features for their platform, I think we'll need a pluggable system instead of `if`s and `switch`es 🙂

pkg/runtime/runtime.go
yevgeny-shnaidman: List is general function that is used by multiple packages.  It think it it better to keep it as is. If you remember we have a task of revising  clients package API.  So if any change is needed will be done during that task

pkg/runtime/runtime.go
yevgeny-shnaidman: This way it is all printed in the same log line, will be hard to understand, no?

pkg/runtime/runtime.go
pmtk: Correct me if I'm wrong, but `List` takes a `client.ObjectList` interface which `v1.SecretList` implements so it should work. Following compiles:
```
var _ client.ObjectList = (*v1.SecretList)(nil)
```

pkg/runtime/runtime.go
yevgeny-shnaidman: yes, we can remove this loop. It was in times when operator was exiting on error, so they wanted to skip restart

yevgeny-shnaidman: fixed

pkg/runtime/runtime.go
yevgeny-shnaidman: fixed

pkg/runtime/runtime.go
qbarrand: I think it's better to log in an atomic way. If we start doing concurrent reconciliations, we don't want two runtime informations being interleaved. I believe this is also better when logs are collected by a cluster-wide sink.

pkg/runtime/runtime.go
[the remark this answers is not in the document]
yevgeny-shnaidman: ok, not sure that everyting can be pluggable, but i will remove the platform

pkg/runtime/runtime.go
yevgeny-shnaidman: fixed

pkg/runtime/runtime.go
yevgeny-shnaidman: fixed

PR #119, opened 2022-02-28
Cache DTK last layer info

pacevedom:
Every reconciliation loop starts by grabbing the DTK image, extracting last layer and then look up some metadata in it. Since we are reading from containers in the registry this information is not going to change for the same image, therefore its not needed to check everytime as we know the outcome.
This PR stores the information upon first retrieval and later on shaves 5s off each reconcile loop.

### DIFF
--- a/pkg/upgrade/upgrade.go
+++ b/pkg/upgrade/upgrade.go
@@ -9,8 +9,6 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
 
-	v1 "github.com/google/go-containerregistry/pkg/v1"
-
 	"github.com/openshift-psap/special-resource-operator/pkg/cluster"
 	"github.com/openshift-psap/special-resource-operator/pkg/registry"
 	"github.com/openshift-psap/special-resource-operator/pkg/utils"
@@ -47,13 +45,20 @@ func NewClusterInfo(registry registry.Registry, cluster cluster.Cluster) Cluster
 		log:      zap.New(zap.UseDevMode(true)).WithName(utils.Print("upgrade", utils.Blue)),
 		registry: registry,
 		cluster:  cluster,
+		cache:    make(map[string]*cacheEntry),
 	}
+}
+
+type cacheEntry struct {
+	ImageURL string
+	DTK      registry.DriverToolkitEntry
 }
 
 type clusterInfo struct {
 	log      logr.Logger
 	registry registry.Registry
 	cluster  cluster.Cluster
+	cache    map[string]*cacheEntry
 }
 
 // GetClusterInfo returns a map[full kernel version]NodeVersion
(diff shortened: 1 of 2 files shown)


pkg/upgrade/upgrade.go
```diff
@@ -175,44 +182,47 @@ func (ci *clusterInfo) updateInfo(info map[string]NodeVersion, dtk registry.Driv
...
+		if cached, ok := ci.cache[entry]; ok {
+			dtk = cached.DTK
+			imageURL = cached.ImageURL
+			ci.log.Info("History from cache", "entry", entry, "dtk", dtk, "imageURL", imageURL)
+		} else {
+			var (
+				err   error
+				layer v1.Layer
+			)
+			layer, err = ci.registry.LastLayer(ctx, entry)
```
qbarrand, line 197:
```suggestion
			layer, err := ci.registry.LastLayer(ctx, entry)
```

pkg/upgrade/upgrade.go
```diff
@@ -175,44 +182,47 @@ func (ci *clusterInfo) updateInfo(info map[string]NodeVersion, dtk registry.Driv
...
+			if err != nil {
+				return nil, err
+			}
+			if layer == nil {
+				continue
+			}
```
qbarrand, line 203: Can a layer ever be `nil` if `err == nil`?

pkg/upgrade/upgrade.go
```diff
@@ -47,13 +47,20 @@ func NewClusterInfo(registry registry.Registry, cluster cluster.Cluster) Cluster
...
+type cacheEntry struct {
+	ImageURL string
+	DTK      registry.DriverToolkitEntry
+}
+
 type clusterInfo struct {
 	log      logr.Logger
 	registry registry.Registry
 	cluster  cluster.Cluster
+	cache    map[string]cacheEntry
```
qbarrand, line 63: Shall we make it `map[string]*cacheEntry`? Otherwise I think accessing the map entry will result in a copy of the `cacheEntry` instance.

pkg/upgrade/upgrade.go
```diff
@@ -175,44 +182,47 @@ func (ci *clusterInfo) updateInfo(info map[string]NodeVersion, dtk registry.Driv
...
-		if layer, err = ci.registry.LastLayer(ctx, imageURL); layer == nil {
-			return nil, fmt.Errorf("cannot extract last layer for DTK from %s: %w", imageURL, err)
-		}
-
-		dtk, err := ci.registry.ExtractToolkitRelease(layer)
-		if err != nil {
-			return nil, err
+		var dtk registry.DriverToolkitEntry
+		var imageURL string
+		if cached, ok := ci.cache[entry]; ok {
```
qbarrand, line 188: If we switch to `*cacheEntry` values in the map, we can do:
```suggestion
		if cached := ci.cache[entry]; cached != nil {
```

qbarrand (commented):
Making the reconciliation faster, one PR at a time. Thank you!

pkg/upgrade/upgrade.go
pacevedom: Good catch, this one depends on https://github.com/openshift/special-resource-operator/pull/114, but will assume it is already there.

pkg/upgrade/upgrade.go
```diff
@@ -175,44 +180,41 @@ func (ci *clusterInfo) updateInfo(info map[string]NodeVersion, dtk registry.Driv
...
-			return nil, err
+		var dtk registry.DriverToolkitEntry
+		var imageURL string
+		if cached := ci.cache[entry]; cached != nil {
+			dtk = cached.DTK
+			imageURL = cached.ImageURL
+			ci.log.Info("History from cache", "entry", entry, "dtk", dtk, "imageURL", imageURL)
+		} else {
+			var err error
+			layer, err := ci.registry.LastLayer(ctx, entry)
```
qbarrand, line 192:
```suggestion
			layer, err := ci.registry.LastLayer(ctx, entry)
```

pacevedom: I would need to declare the error somewhere below, as lines 202 and 204 use the variables defined in lines 187 and 188. Dont want to hide them by using `:=`.

pkg/upgrade/upgrade.go
```diff
@@ -47,13 +45,20 @@ func NewClusterInfo(registry registry.Registry, cluster cluster.Cluster) Cluster
...
+type cacheEntry struct {
+	ImageURL string
+	DTK      registry.DriverToolkitEntry
+}
+
 type clusterInfo struct {
 	log      logr.Logger
 	registry registry.Registry
 	cluster  cluster.Cluster
+	cache    map[string]*cacheEntry
```
yevgeny-shnaidman, line 61: just one small comment: it means that we are going to keep all the DTK version for all the upgrade that has been executed on the cluster. Potentially the cache can grow to hundreds of entries

qbarrand: Won't the masters likely be restarted at each upgrade as well?

pacevedom: This should be capped by the amount of z-streams. Any minor/major upgrade to the cluster would also trigger an upgrade to SRO, so it should not grow infinitely. Also, any node drain and/or pod reschedule would clean this out.

qbarrand:
@pacevedom also wondering if we could unit test this.

PR #121, opened 2022-03-01
Makefile fixes for: no rule 'docker-push'

ukalifon:
Remove calls to target docker-push, which doesn't exist


ukalifon:
Closig. See PR #122 instead.

PR #122, opened 2022-03-01
Bug 2060007: Makefile fixes for: no rule 'docker-push'

ukalifon:
Remove calls to target docker-push, which doesn't exist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,6 +49,11 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
+# If CONTAINER_COMMAND wasn't specified, default to podman
+ifeq (,$(CONTAINER_COMMAND))
+CONTAINER_COMMAND=podman
+endif
+
 # Setting SHELL to bash allows bash commands to be executed by recipes.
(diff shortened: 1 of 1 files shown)


pacevedom:
Not sure, but are we allowed to push to 4.10 without a BZ?

qbarrand:
Yes, I would first merge this into `master`, then cherry-picked into the release branch if possible.
Also, could we merge this change upstream first?

ukalifon:
@qbarrand @yevgeny-shnaidman It doesn't make sense to merge this into master first, because the Makefile there is vastly different from this one, and in master the container command is parameterized. So instead, I fixed my PR to have the container command parameterized in this branch as well, pretty similar to how it is in master.

ukalifon:
BZ is here: https://bugzilla.redhat.com/show_bug.cgi?id=2060007

qbarrand:
@ukalifon please prepend the PR's title with `Bug 2060007:` and then comment `/bugzilla refresh`.

/lgtm

PR #123, opened 2022-03-02
Remove wait from poll package

pacevedom:
Remove wait calls from poll package.

Reconciliation loop goes through all yaml files in a helm chart, each of them having the following waiting pattern:
- Wait 5s before doing any checks.
- Check if the resource is available (whatever that is for different resources).
- If it is, return error-free and continue with the next file.
- If not, retry after 5 additional seconds.
- Repeat until 30s mark is reached and return an error.

This means that each file will introduce a significant delay in the loop, preventing other SRs to progress and, in general, taking too long to execute each one.

This PR drops all waiting and replaces them with immediate checks with reconcile errors to allow concurrency. Results of the change are high speedups when reconciling, plus improved responsiveness.

### DIFF
--- a/controllers/specialresource_controller.go
+++ b/controllers/specialresource_controller.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"context"
 	"os"
+	"time"
 
 	"github.com/go-logr/logr"
 	buildv1 "github.com/openshift/api/build/v1"
@@ -31,6 +32,7 @@ import (
 	storagev1 "k8s.io/api/storage/v1"
 	k8sruntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/workqueue"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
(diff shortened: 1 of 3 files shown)


controllers/specialresource_controller.go
```diff
@@ -159,6 +166,7 @@ func (r *SpecialResourceReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			Owns(&v1.Secret{}).
 			WithOptions(controller.Options{
 				MaxConcurrentReconciles: 1,
+				RateLimiter:             workqueue.NewItemExponentialFailureRateLimiter(minDelay, maxDelay),
```
qbarrand, line 169: I like this, but is this aligned with a standard within OCP? Should it maybe be user-configurable?

yevgeny-shnaidman: Why not use a a requeue with timeout in specific places, instead of changing the default values?

pacevedom: @qbarrand No standard as far as I know. Not sure about making this configurable, as it can render the operator unresponsive/slow.
@yevgeny-shnaidman Right now any error falls under these intervals, so that it gets quickly reconciled again. I cant see a use case where we would need to wait longer than others with no prior knowledge (for example, knowing in advance that certain resources take at least X seconds). Besides, default values are maybe too tight on both ends. The minimum is 100ms while it maxes out at 1000s. Duplicating wait on each failure escalates pretty quickly to long queueing.

bthurber (approved):
/lgtm

PR #124, opened 2022-03-02
Use helm libraries as client only

pacevedom:
Shaves around ~10s from each reconciliation loop.

Helm is used to dry run an install to generate all manifests for later apply in an ordered fashion from SRO. Among the options for the install we have `ClientOnly` set to false. If we look at how helm uses this parameter we find (assuming it is set to false):
- Install config runs using `DryRun`, meaning no actual resources will be created.
- When setting `ClientOnly` to false helm uses an actual kubeclient to talk to the cluster.
- CRDs get installed if not using `DryRun`.
- Capabilities and API versions are retrieved from the cluster.
- Capabilities and API versions are passed to render manifests coalescing values.
- Check if resources already exist in the cluster.
- Dry run active, bail out.

Capabilities are a mixture of API versions and kubernetes version. These are used as part of the helm [values](https://github.com/helm/helm/blob/main/pkg/chartutil/values.go#L143) when rendering templates, like `{{ .Release.name }}`, for example.
When using `ClientOnly` this means the API versions are not retrieved from the cluster, and resources are not checked for existence either. Capabilities are filled either from default templates having the supported standard k8s resources, depending on helm version, or provided to the install config via attributes. We chose the latter, as the capabilities only change after upgrades and stay stable for the lifetime of SRO. When creating an instance of the `helmer` package all this data is retrieved for later use.

### DIFF
--- a/main.go
+++ b/main.go
@@ -133,7 +133,7 @@ func main() {
 
 	clusterInfoAPI := upgrade.NewClusterInfo(registry.NewRegistry(kubeClient), clusterAPI)
 	runtimeAPI := runtime.NewRuntimeAPI(kubeClient, clusterAPI, kernelAPI, clusterInfoAPI, proxyAPI)
-	helmerAPI, err := helmer.NewHelmer(creator, helmer.DefaultSettings(), kubeClient)
+	helmerAPI, err := helmer.NewHelmer(creator, helmSettings, kubeClient)
 	if err != nil {
 		setupLog.Error(err, "Unable to setup helmer")
 		os.Exit(1)
--- a/main.go
+++ b/main.go
@@ -133,6 +133,11 @@ func main() {
 
 	clusterInfoAPI := upgrade.NewClusterInfo(registry.NewRegistry(kubeClient), clusterAPI)
 	runtimeAPI := runtime.NewRuntimeAPI(kubeClient, clusterAPI, kernelAPI, clusterInfoAPI, proxyAPI)
+	helmerAPI, err := helmer.NewHelmer(creator, helmer.DefaultSettings(), kubeClient)
+	if err != nil {
+		setupLog.Error(err, "Unable to setup helmer")
+		os.Exit(1)
+	}
 
 	if err = (&controllers.SpecialResourceReconciler{Cluster: clusterAPI,
 		ClusterInfo:            clusterInfoAPI,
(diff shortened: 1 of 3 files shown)


pkg/helmer/helmer.go
```diff
@@ -240,7 +240,7 @@ func (h *helmer) Run(
 	install.DryRun = true
 	install.ReleaseName = ch.Metadata.Name
 	install.Replace = false
-	install.ClientOnly = false
+	install.ClientOnly = true
```
yevgeny-shnaidman, line 243: What about capabilities object? Do we need helm to build based on the cluster?

pacevedom: You just make me realise we can compute capabilities elsewhere and not in the loop. Going to give it a shot, thanks!

pkg/helmer/helmer.go
```diff
@@ -263,13 +286,12 @@ func (h *helmer) Run(
 		return fmt.Errorf("Chart has an unsupported type %s and can not be installed", ch.Metadata.Type)
 	}
 
-	// Pre-install anything in the crd/ directory. We do this before Helm
-	// contacts the upstream server and builds the capabilities object.
-	if crds := ch.CRDObjects(); !install.ClientOnly && !install.SkipCRDs && len(crds) > 0 {
+	// Pre-install anything in the crd/ directory.
+	if crds := ch.CRDObjects(); len(crds) > 0 {
```
yevgeny-shnaidman, line 290: do we need to install CRDs before in install.Run? can it be moved to a after install?

pacevedom: Actually, install.Run only renders manifests. The InstallCRDs function is copied from the one inside helm packages, as `Run` is doing that when not in dry run mode.

pacevedom: Let me do it alongside rebasing, as it poses no difference to the current approach.

main.go
```diff
@@ -133,7 +134,25 @@ func main() {
 
 	clusterInfoAPI := upgrade.NewClusterInfo(registry.NewRegistry(kubeClient), clusterAPI)
 	runtimeAPI := runtime.NewRuntimeAPI(kubeClient, clusterAPI, kernelAPI, clusterInfoAPI, proxyAPI)
-	helmerAPI, err := helmer.NewHelmer(creator, helmSettings, kubeClient)
+
+	helmDiscoveryClient, err := helmSettings.RESTClientGetter().ToDiscoveryClient()
```
yevgeny-shnaidman, line 138: I think it is not a very pretty code, when the inner working of the package are initiated in the main, and all this is done just for unit-test. Maybe we can add a inner function to the helmer package, so that unit-test can initiate the Helmer package for unit-test as it sees fit:
func NewHelmer() {
   /* init dc, kubeVersion, api Version
   newHelmer(creator resource.Creator, settings *cli.EnvSettings, version *version.Info, apiVersions chartutil.VersionSet, kubeClient clients.ClientsInterface)
}

and unit test can initialize dc, kubeVersion, api Version and call newHelmer

pacevedom: Finally went with your approach, mocking was working but tests were so convoluted it didnt look easy to change something.

PR #125, opened 2022-03-03
Introduce work item and reduce global vars

pmtk:
PR focuses on removing couple of vars from SpecialResourceReconciler. These variables were used like global vars making it hard to reason about the code and possibly introducing bugs (like using object from previous reconcile).

Summary:
* Functions taking a `*SpecialResourceReconciler` as a parameter are now methods of that struct.
* `SpecialResourceReconciler` no longer has SpecialResource, Chart, Values members.
* New struct `WorkItem` was created containing above data as needed. `WorkItem` is passed as a parameter to functions.
* Package `state` was removed. Its `CurrentName` is incorporated as parameter to relevant functions.
* `runtime` package now offers a fresh object whenever requested. RunInfo is no longer a long-lived object


controllers/workitem.go
```diff
@@ -0,0 +1,37 @@
...
+	srov1beta1 "github.com/openshift-psap/special-resource-operator/api/v1beta1"
+	"github.com/openshift-psap/special-resource-operator/pkg/runtime"
+	"github.com/openshift-psap/special-resource-operator/pkg/utils"
+	"helm.sh/helm/v3/pkg/chart"
+)
+
+// WorkItem stores values required for current reconciliation
+type WorkItem struct {
+	// Log is a logger dedicated for specific SpecialResource constructed with its NamespacedName.
+	Log logr.Logger
```
yevgeny-shnaidman, line 14: why do we need log i the WorkItem? Each package has a log of its own

controllers/label_nodes.go
```diff
@@ -4,44 +4,36 @@ import (
...
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 )
 
 // If resource available, label the nodes according to the current state
 // if e.g driver-container ready -> specialresource.openshift.io/driver-container:ready
-func (r *SpecialResourceReconciler) labelNodesAccordingToState(ctx context.Context, nodeSelector map[string]string) error {
+func (r *SpecialResourceReconciler) labelNodesAccordingToState(ctx context.Context, log logr.Logger, nodeSelector map[string]string, key string) error {
 
-	nodeList, err := r.KubeClient.GetNodesByLabels(ctx, r.specialresource.Spec.NodeSelector)
+	nodeList, err := r.KubeClient.GetNodesByLabels(ctx, nodeSelector)
```
yevgeny-shnaidman, line 15: General question: why do we need to label the nodes with labels per CR? It is not used anywhere, and we do have logs/metrics that supply the same info the user. @qbarrand , @pmtk @pacevedom wdyt?

controllers/workitem.go
qbarrand: This might be to always have a logger that has a field set to the SpecialResource's name.

controllers/label_nodes.go
qbarrand: Can't think of a valid reason either. If you want to know on how many nodes your SpecialResource is present, you can always check the DaemonSet's status.

pacevedom: I always wondered why we needed that. Its written and then not read anywhere.

yevgeny-shnaidman: so lets open a Jira to remove it. Agree?

controllers/workitem.go
yevgeny-shnaidman: i think that in that case we should use context

yevgeny-shnaidman: then maybe what we need to do is to remove the log variable from all the packages, and only use the log in the Context. It also means that context needs to be passed to most of the APIs of the packages

pmtk: > This might be to always have a logger that has a field set to the SpecialResource's name
That is exactly the reason. When creating a new logger using `.WithName()` the logger's name will grow: `ping-pong.cert-manager` - I think this might be useful for tracking what action we perform for which object.

> then maybe what we need to do is to remove the log variable from all the packages, and only use the log in the Context. It also means that context needs to be passed to most of the APIs of the packages
I think passing logger to package's methods is good idea if we'd have concurrent reconciles. Passing it via context is nice idea, but I'm not sure about extra boilerplate (to unpack it every time) and kind of implicit contract (there's no compile-time check to ensure you're passing a logger there)

yevgeny-shnaidman: unpack logger is supposed to be a very quick method

pmtk: I didn't mean the performance - rather the extra code. Nonetheless, I think we can investigate that but maybe in separate PR?

yevgeny-shnaidman: yes, lets open it right now, so we don't forget :)

PR #126, opened 2022-03-04
Properly determine kernel affinity in controller

pacevedom:


controllers/resources.go
```diff
@@ -19,6 +19,10 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
+var (
+	affineRegex = regexp.MustCompile(`\n\s+specialresource\.openshift\.io\/kernel\-affine`)
```
qbarrand, line 23: I'm missing if the `\n` is there on purpose. If not, how about:
```suggestion
	affineRegex = regexp.MustCompile(`^\s+specialresource\.openshift\.io/kernel-affine:.+$`)
```
I think some escapes might be redundant.

pacevedom: It was on purpose, but you got in my head the multiline flag that I forgot in the first place. Thanks!

PR #132, opened 2022-03-09
MGMT-9551 - extend Regsitry API to allow querring different image layers

yevgeny-shnaidman:
This PR adds 2 APIs that:
1) return digest of all the layers in the image, in addition to
   repo and repo auth
2) return v1Layer based on a provided repo,digest and auth

### DIFF
--- a/pkg/registry/registry.go
+++ b/pkg/registry/registry.go
@@ -46,6 +46,8 @@ type Registry interface {
 	LastLayer(context.Context, string) (v1.Layer, error)
 	ExtractToolkitRelease(v1.Layer) (*DriverToolkitEntry, error)
 	ReleaseManifests(v1.Layer) (string, string, error)
+	GetLayersDigests(context.Context, string) (string, []string, []crane.Option, error)
+	GetLayerByDigest(string, string, []crane.Option) (v1.Layer, error)
 }
 
 func NewRegistry(kubeClient clients.ClientsInterface) Registry {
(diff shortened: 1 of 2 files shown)


pkg/registry/registry.go
```diff
@@ -302,6 +265,56 @@ func (r *registry) ReleaseManifests(layer v1.Layer) (string, string, error) {
...
+		return "", nil, nil, err
+	}
+
+	var repo string
+
+	if hash := strings.Split(entry, "@"); len(hash) > 1 {
+		repo = hash[0]
+	} else if tag := strings.Split(entry, ":"); len(tag) > 1 {
+		repo = tag[0]
+	}
```
qbarrand, line 285: If `entry` does not contain neither `@` nor `:`, then `repo` will remain empty. Should we then set it to `entry`?
```suggestion
	repo := entry

	if hash := strings.Split(repo, "@"); len(hash) > 1 {
		repo = hash[0]
	} else if tag := strings.Split(repo, ":"); len(tag) > 1 {
		repo = tag[0]
	}
```

pkg/registry/registry.go
```diff
@@ -302,6 +265,56 @@ func (r *registry) ReleaseManifests(layer v1.Layer) (string, string, error) {
...
+	digests := make([]string, len(layers))
+	for i, layer := range layers {
+		digests[i] = layer.(map[string]interface{})["digest"].(string)
+	}
+
+	return repo, digests, registryAuths, nil
+}
+
+func (r *registry) GetLayerByDigest(repo string, digest string, auth []crane.Option) (v1.Layer, error) {
```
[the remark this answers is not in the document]
yevgeny-shnaidman: This will be used by the Preflight: we want to get a DTK info from an drvier container image, but most of the chances it will be stored on the last layer

pkg/registry/registry.go
```diff
@@ -302,6 +265,56 @@ func (r *registry) ReleaseManifests(layer v1.Layer) (string, string, error) {
...
+	if auth.Auth != "" {
+		registryAuths = append(registryAuths, crane.WithAuth(authn.FromConfig(authn.AuthConfig{Username: auth.Email, Auth: auth.Auth})))
+	}
+
+	manifest, err := crane.Manifest(entry, registryAuths...)
+	if err != nil {
+		return "", nil, nil, err
+	}
+
+	release := unstructured.Unstructured{}
```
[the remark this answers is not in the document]
yevgeny-shnaidman: this is the only place we use it, so i think we can leave it as is, and not create a specific type that we need to maintain just for those 2 lines, but if you really insist i can change it :)

pkg/registry/registry.go
yevgeny-shnaidman: i don't think so. we always assume that there is either a tag or a hash in the image. leaving the repo empty will result in failure to get the image, which i think is ok. Another option is to return an error in case repo is empty

qbarrand: Returning an error sounds good to me.

pkg/registry/registry.go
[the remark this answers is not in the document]
qbarrand: I have a bias against `unstructured` 🙂 but will tolerate that one if you prefer not to change it.

pkg/registry/registry.go
yevgeny-shnaidman: fixed

PR #133, opened 2022-03-09
MGMT-9548: adding API to get DTK info based on DTK image

yevgeny-shnaidman:
This PR adds an API that allows to get an DTK info (full kernel, rt kernel etc')
encoded inside the DTK image

### DIFF
--- a/pkg/upgrade/upgrade.go
+++ b/pkg/upgrade/upgrade.go
@@ -27,6 +27,7 @@ type NodeVersion struct {
 
 type ClusterInfo interface {
 	GetClusterInfo(context.Context, *corev1.NodeList) (map[string]NodeVersion, error)
+	GetDTKData(ctx context.Context, imageURL string) (*registry.DriverToolkitEntry, error)
 }
 
 func NewClusterInfo(reg registry.Registry, cluster cluster.Cluster) ClusterInfo {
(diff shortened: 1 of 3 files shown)


pkg/upgrade/upgrade.go
```diff
@@ -146,8 +147,22 @@ func (ci *clusterInfo) driverToolkitVersion(ctx context.Context, dtkImages []str
...
+	// we're going only to update the struct with DTK information on
+	// running kernels and not on all that are found.
+	// We could have many entries with DTKs that are from an old update
+	// The objects that are kernel affine should only be replicated
+	// for valid kernels.
+	return ci.updateInfo(info, *dtk, imageURL)
+}
 
-	if dtk = ci.cache[imageURL]; dtk != nil {
+func (ci *clusterInfo) GetDTKData(ctx context.Context, imageURL string) (*registry.DriverToolkitEntry, error) {
```
pacevedom, line 163: I guess this will be used for the preflight checks as well, right?

yevgeny-shnaidman: yes

pkg/upgrade/upgrade_test.go
```diff
@@ -350,3 +352,84 @@ var _ = Describe("ClusterInfo", func() {
 		})
 	})
 })
+
+var _ = Describe("GetDTKData", func() {
```
pacevedom, line 356: Since we now have a dedicated function, could we remove the caching tests for `GetClusterInfo`?

yevgeny-shnaidman: yes, fixed

pacevedom (approved):
/lgtm

PR #138, opened 2022-03-15
Hub and spokes ACM functionality

pacevedom:
Includes:
- New CRD
- New controller
- New logic

Everything (except for filter package, which is modified to accomodate both controllers) is completely separated from previous CRD handling.


qbarrand:
/hold
Just to prevent the PR from being merged while I am reviewing it 🙂

charts/example/acm-ice-0.0.1/templates/0003-policy.yaml
```diff
@@ -0,0 +1,70 @@
+apiVersion: policy.open-cluster-management.io/v1
```
pmtk, line 1: Nit: The 0002 file is missing, maybe renaming the files  0001-0003 would avoid potential confusion?

config/rbac/role.yaml
```diff
@@ -274,6 +274,17 @@ rules:
   - patch
   - update
   - watch
+- apiGroups:
+  - apps.open-cluster-management.io
```
pmtk, line 278: Can we avoid placing ACM related stuff here? Would prepending RBAC manifests to SRM work?

manifests/0001_customresourcedefinition_specialresourcemodules.sro.openshift.io.yaml
```diff
@@ -0,0 +1,209 @@
+apiVersion: apiextensions.k8s.io/v1
```
pmtk, line 1: These `manifests/` files seems fishy. They duplicate numbers in different order.
Shouldn't be a diff merged into existing manifests (like new CRD)?

pkg/watcher/watcher.go
```diff
@@ -0,0 +1,359 @@
...
+	}
+	if ok {
+		w.log.Info("ok", "paths", paths, "name", o.GetName())
+		for _, path := range paths {
+			// Get current value of property at path
+			vals, err := GetJSONPath(path, *unstr)
+			if err != nil {
+				w.log.Error(err, "failed to obtain a value", "path", path)
+			}
+			w.log.Info("obtained values @ path", "path", path, "values", vals)
```
pmtk, line 289: I remember this function's logs being quite noisy. If we'd have a leveled logger (I think it's only a matter of initializing it with an arg rather than defaulting to `dev` mode) then these logs could be level > 1.

controllers/specialresourcemodule_controller.go
```diff
@@ -0,0 +1,535 @@
...
+	full, major, minor := res[0], res[1], res[2]
+	var imageURL string
+	{
+		transport, _ := transport.HTTPWrappersForConfig(
+			&transport.Config{
+				UserAgent: rest.DefaultKubernetesUserAgent() + "(release-info)",
+			},
+			http.DefaultTransport,
+		)
+		client := &http.Client{Transport: transport}
```
pmtk, line 156: This is shadowing `"sigs.k8s.io/controller-runtime/pkg/client"`

controllers/specialresourcemodule_controller.go
```diff
@@ -0,0 +1,535 @@
...
+		Payload string `json:"payload"`
+	}
+	type versionGraph struct {
+		Nodes []versionNode `json:"nodes"`
+	}
+	res := versionRegex.FindStringSubmatch(entry)
+	full, major, minor := res[0], res[1], res[2]
+	var imageURL string
+	{
+		transport, _ := transport.HTTPWrappersForConfig(
```
pmtk, line 150: This is shadowing `"k8s.io/client-go/transport"`

manifests/0001_customresourcedefinition_specialresourcemodules.sro.openshift.io.yaml
pacevedom: Forgot to generate new manifests

controllers/specialresourcemodule_controller.go
```diff
@@ -0,0 +1,535 @@
...
+
+	if resource.Spec.Namespace != "" {
+		add := []byte(resource.Spec.Namespace)
+		ns = append(ns, add...)
+	} else {
+		resource.Spec.Namespace = resource.Name
+		add := []byte(resource.Spec.Namespace)
+		ns = append(ns, add...)
+	}
+	return r.Creator.CreateFromYAML(ctx, ns, false, &resource, resource.Name, "", nil, "", "", SRMOwnedLabel)
```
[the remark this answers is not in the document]
pacevedom: The `wait` annotation is not handled if using objects instead of yaml files, as its SRO taking care of that.

controllers/specialresourcemodule_controller.go
pacevedom: Coming straight from https://github.com/openshift/oc, while it is true there is shadowing its only in this scope. Will change it though because it might be confusing.

controllers/specialresourcemodule_controller.go
pacevedom: Same comment.

controllers/specialresourcemodule_controller.go
```diff
@@ -0,0 +1,535 @@
...
+			&transport.Config{
+				UserAgent: rest.DefaultKubernetesUserAgent() + "(release-info)",
+			},
+			http.DefaultTransport,
+		)
+		client := &http.Client{Transport: transport}
+		u, _ := url.Parse("https://api.openshift.com/api/upgrades_info/v1/graph")
+		for _, stream := range []string{"fast", "stable", "candidate"} {
+			u.RawQuery = url.Values{"channel": []string{fmt.Sprintf("%s-%s.%s", stream, major, minor)}}.Encode()
+			if err := func() error {
```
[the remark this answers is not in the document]
pacevedom: Coming straight from https://github.com/openshift/oc, I can extract it from the function.

charts/global-values.yaml
```diff
@@ -13,6 +13,17 @@ specialresource:
...
+specialResourceModule:
+  metadata:
+    name: "name"
+  spec:
+    namespace: "namespace"
+    driverContainer:
+      source:
+        git:
+          ref: "ref"
+          uri: "uri"
```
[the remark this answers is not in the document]
pacevedom: This is for helm lint make target, otherwise there are missing values and it fails.

config/rbac/role.yaml
pacevedom: Not sure what you mean by prepending to SRM, you mean the CR?

charts/example/acm-ice-0.0.1/templates/0001-buildconfig.yaml
```diff
@@ -0,0 +1,98 @@
...
+    set -eu
+
+    # unload in-tree driver
+    rmmod ice || true
+
+    # load out-of-tree driver
+    insmod /ice-driver/ice.ko
+
+    echo "Loaded out-of-tree ICE"
+    lsmod | grep ice || true
```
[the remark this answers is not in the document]
pacevedom: It does fail, there is a `set -e` at the top. This is just informative, we could as well remove it.

charts/example/acm-ice-0.0.1/templates/0003-policy.yaml
```diff
@@ -0,0 +1,70 @@
...
+                     files:
+                       - contents:
+                           source: 'data:text/plain;charset=us-ascii;base64,IyEvYmluL2Jhc2gKc2V0IC1ldQoKQUNUSU9OPSQxOyBzaGlmdApJTUFHRT0kMTsgc2hpZnQKS0VSTkVMPWB1bmFtZSAtcmAKCnBvZG1hbiBwdWxsIC0tYXV0aGZpbGUgL3Zhci9saWIva3ViZWxldC9jb25maWcuanNvbiAke0lNQUdFfToke0tFUk5FTH0gMj4mMQoKbG9hZF9rbW9kcygpIHsKCiAgICBwb2RtYW4gcnVuIC1pIC0tcHJpdmlsZWdlZCAtdiAvbGliL21vZHVsZXMvJHtLRVJORUx9L2tlcm5lbC9kcml2ZXJzLzovbGliL21vZHVsZXMvJHtLRVJORUx9L2tlcm5lbC9kcml2ZXJzLyAke0lNQUdFfToke0tFUk5FTH0gbG9hZC5zaAp9CnVubG9hZF9rbW9kcygpIHsKICAgIHBvZG1hbiBydW4gLWkgLS1wcml2aWxlZ2VkIC12IC9saWIvbW9kdWxlcy8ke0tFUk5FTH0va2VybmVsL2RyaXZlcnMvOi9saWIvbW9kdWxlcy8ke0tFUk5FTH0va2VybmVsL2RyaXZlcnMvICR7SU1BR0V9OiR7S0VSTkVMfSB1bmxvYWQuc2gKfQoKY2FzZSAiJHtBQ1RJT059IiBpbgogICAgbG9hZCkKICAgICAgICBsb2FkX2ttb2RzCiAgICA7OwogICAgdW5sb2FkKQogICAgICAgIHVubG9hZF9rbW9kcwogICAgOzsKICAgICopCiAgICAgICAgZWNobyAiVW5rbm93biBjb21tYW5kLiBFeGl0aW5nLiIKICAgICAgICBlY2hvICJVc2FnZToiCiAgICAgICAgZWNobyAiIgogICAgICAgIGVjaG8gImxvYWQgICAgICAgIExvYWQga2VybmVsIG1vZHVsZShzKSIKICAgICAgICBlY2hvICJ1bmxvYWQgICAgICBVbmxvYWQga2VybmVsIG1vZHVsZShzKSIKICAgICAgICBleGl0IDEKZXNhYwo='
```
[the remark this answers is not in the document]
pacevedom: The script that the systemd service launches:
```
#!/bin/bash
set -eu

ACTION=$1; shift
IMAGE=$1; shift
KERNEL=`uname -r`

podman pull --authfile /var/lib/kubelet/config.json ${IMAGE}:${KERNEL} 2>&1

load_kmods() {

    podman run -i --privileged -v /lib/modules/${KERNEL}/kernel/drivers/:/lib/modules/${KERNEL}/kernel/drivers/ ${IMAGE}:${KERNEL} load.sh
}
unload_kmods() {
    podman run -i --privileged -v /lib/modules/${KERNEL}/kernel/drivers/:/lib/modules/${KERNEL}/kernel/drivers/ ${IMAGE}:${KERNEL} unload.sh
}

case "${ACTION}" in
    load)
        load_kmods
    ;;
    unload)
        unload_kmods
    ;;
    *)
        echo "Unknown command. Exiting."
        echo "Usage:"
        echo ""
        echo "load        Load kernel module(s)"
        echo "unload      Unload kernel module(s)"
        exit 1
esac
```

charts/example/acm-ice-0.0.1/acm-ice.yaml
```diff
@@ -0,0 +1,29 @@
...
+metadata:
+  name: acm-ice
+spec:
+  namespace: acm-ice
+  chart:
+    name: acm-ice
+    version: 0.0.1
+    repository:
+      name: chart
+      url: file:///charts/example
```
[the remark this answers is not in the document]
pacevedom: Maybe swapping this with `cm`, but then we need to host it in a different directory. This is just for example purposes.

config/rbac/role.yaml
pmtk: Yes, packing necessary RBACs into a yaml file containing SRM/CR manifest

pacevedom:
PTAL @qbarrand @yevgeny-shnaidman @pmtk @ybettan @enriquebelarte

ybettan:
@pacevedom Can you please squash all the commits?

ybettan:
@pacevedom Thanks for squashing.
The commit message still is a long list of all older commits titles.
Can we please make it a logical commit message describing the change? (Same as you did fo the PR description).

PR #139, opened 2022-03-17
Fix csplit for manifests-gen

pacevedom:


ybettan:
/lgtm
/hold
@pacevedom please unhold if we still need that merge.

PR #142, opened 2022-03-18
Introduce standard node affinity support. Replace NFD label scheduling

pacevedom:
Remove NFD label set when using affinity annotations in recipes. Removes all knowledge of NFD labels from the operator, using only standard resources to determine where to schedule workloads that need a specific kernel.
Also addresses problems with skew when upgrading nodes if using NFD.


yevgeny-shnaidman:
@pacevedom  what is the reason for this PR? Not sure i understand why we need to set the names affinity

pacevedom:
> @pacevedom what is the reason for this PR? Not sure i understand why we need to set the names affinity

This is the second part of removing NFD dependency, also useful for 2.0 as it will have a similar/same mechanism. Using NFD labels might be convenient in some situations but gets you into additional problems like dependencies, configuration, skew when ugprading, etc.
This PR solves all that by using standard k8s ways of scheduling. Now NFD labels are entirely unknown to SRO, only relevant to recipes, if they have such needs.

pkg/upgrade/upgrade.go
```diff
@@ -81,12 +80,19 @@ func (ci *clusterInfo) nodeVersionInfo(nodeList *corev1.NodeList) (map[string]No
...
-			OSMajor:        "rhel" + osMajor,
-			OSMajorMinor:   "rhel" + osVersion,
+		data, ok := info[kernelFullVersion]
+		if !ok {
+			data = NodeVersion{
+				OSVersion:      osVersion,
+				ClusterVersion: clusterVersion,
+				OSMajor:        "rhel" + osMajor,
+				OSMajorMinor:   "rhel" + osVersion,
+				Names:          []string{node.Name},
```
yevgeny-shnaidman, line 90: why do we need the names of the nodes? we are still keeping the kernel full version, and we are getting it from the node, and not from the NFD labels. So why can't we just keep using it for affinity?

pacevedom: Using `nodeAffinity` requires either labels or supported fields, and none of these are able to match against `status.kernelVersion` in the nodes. Therefore we need something that we can map to expressions/fields, hence the names.

internal/resourcehelper/helper.go
```diff
@@ -110,36 +110,74 @@ func (rh *resourceHelper) UpdateResourceVersion(req *unstructured.Unstructured,
 
 func (rh *resourceHelper) SetNodeSelectorTerms(obj *unstructured.Unstructured, terms map[string]string) error {
 	switch obj.GetKind() {
-	case "DaemonSet", "Deployment", "Statefulset": // TODO(qbarrand) should this be StatefulSet?:
-		if err := rh.nodeSelectorTerms(terms, obj, "spec", "template", "spec", "nodeSelector"); err != nil {
+	case "DaemonSet", "Deployment", "StatefulSet":
```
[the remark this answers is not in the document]
pacevedom: Its weird indeed. There is a `nodeSelector` at CRD level that is applied to all resources in the helm chart. Then you can have `nodeSelector` per resource if you cofigure them like that. The CRD option is maybe too coarse to do such a thing, as it applies to `BuildConfig`, for example, which is not really needed.

pacevedom:
/hold
Need to test something with upgrades.

ybettan:
/retest
@pacevedom can you please squash the commits?

pacevedom:
PTAL @qbarrand @yevgeny-shnaidman @ybettan @pmtk @enriquebelarte

ybettan:
Can you please update the commit message to be consistent with the PR description?

pacevedom:
Wait until v2 comes in.

bthurber:
Recommend moving this to v2.

PR #144, opened 2022-03-21
Removing all helm commands from the build process.

ybettan:
Those helm commands are needed to manipulate helm charts used by the
`file:///...` url, which was agreed by partners to be removed.

The build process of OCP (including SRO in it) is done in a disconnected
env, therefore, we cannot install the `helm` package during the build
process .

Also removing partners charts.


ybettan:
/cc @qbarrand 
Will create a parallel PR to u/s once this one is final and merged.

ybettan:
Steps:
1. Make it possible to build SRO in a disconnected env. [done]
2. Remove `file:///` for referencing helm charts in the recipes. [wip]

ybettan:
Here is the [PR](https://github.com/openshift/release/pull/27139) to remove `helm-lint` job.

Dockerfile
```diff
@@ -23,9 +22,10 @@ COPY cmd/ cmd/
 COPY controllers/ controllers/
 COPY internal/ internal/
 COPY pkg/ pkg/
+#FIXME:ybettan remove?
```
qbarrand, line 25: Yes 🙂

Dockerfile
```diff
@@ -37,7 +37,8 @@ ENV HELM_PLUGINS /opt/helm-plugins
 
 COPY --from=builder /workspace/manager /manager
 COPY --from=builder /workspace/helm-plugins ${HELM_PLUGINS}
-COPY --from=builder /workspace/build/charts /charts
+#FIXME:ybettan remove?
```
qbarrand, line 40: Yes 🙂

Dockerfile
ybettan: done

PR #145, opened 2022-03-21
MGMT-9690: add OS configuration extraction from release image

yevgeny-shnaidman:
PR changes:
1) add ReleaseImageMachineOSConfig API - get the OS data (major/minor etc'), need by preflight
2) change ReleaseManifests API - no need to return version, not in use
3) refactor function, add common helper function getHeaderFromLayer

### DIFF
--- a/pkg/registry/registry.go
+++ b/pkg/registry/registry.go
@@ -5,6 +5,7 @@ import (
 	"compress/gzip"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/url"
@@ -15,7 +16,6 @@ import (
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/openshift/special-resource-operator/pkg/clients"
 	"github.com/openshift/special-resource-operator/pkg/utils"
-	"github.com/pkg/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
@@ -38,7 +38,8 @@ type DriverToolkitEntry struct {
 type Registry interface {
 	LastLayer(context.Context, string) (v1.Layer, error)
 	ExtractToolkitRelease(v1.Layer) (*DriverToolkitEntry, error)
-	ReleaseManifests(v1.Layer) (string, string, error)
+	ReleaseManifests(v1.Layer) (string, error)
+	ReleaseImageMachineOSConfig(layer v1.Layer) (string, error)
 	GetLayersDigests(context.Context, string) (string, []string, []crane.Option, error)
 	GetLayerByDigest(string, string, []crane.Option) (v1.Layer, error)
 }
@@ -76,7 +77,7 @@ func (r *registry) registryFromImageURL(image string) (string, error) {
 func (r *registry) getImageRegistryCredentials(ctx context.Context, registry string) (dockerAuth, error) {
 	s, err := r.kubeClient.GetSecret(ctx, pullSecretNamespace, pullSecretName, metav1.GetOptions{})
 	if err != nil {
-		return dockerAuth{}, errors.Wrap(err, "could not retrieve pull secrets")
+		return dockerAuth{}, fmt.Errorf("could not retrieve pull secrets, [%w]", err)
 	}
 
 	pullSecretData, ok := s.Data[pullSecretFileName]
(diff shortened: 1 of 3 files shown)


pkg/registry/registry.go
```diff
@@ -110,147 +111,83 @@ func (r *registry) LastLayer(ctx context.Context, entry string) (v1.Layer, error
...
-			dtk.OSVersion = entry
+	dtk.KernelFullVersion = entry
 
-			return dtk, err
-		}
+	entry, _, err = unstructured.NestedString(obj.Object, "RT_KERNEL_VERSION")
+	if err != nil {
+		return nil, err
+	}
+	dtk.RTKernelFullVersion = entry
```
qbarrand, line 129:
```suggestion
	dtk.RTKernelFullVersion, _, err = unstructured.NestedString(obj.Object, "RT_KERNEL_VERSION")
	if err != nil {
		return nil, err
	}
```

pkg/registry/registry.go
```diff
@@ -110,147 +111,83 @@ func (r *registry) LastLayer(ctx context.Context, entry string) (v1.Layer, error
...
 
+	entry, _, err = unstructured.NestedString(obj.Object, "RHEL_VERSION")
+	if err != nil {
+		return nil, err
 	}
+	dtk.OSVersion = entry
```
qbarrand, line 135:
```suggestion
	dtk.OSVersion, _, err = unstructured.NestedString(obj.Object, "RHEL_VERSION")
	if err != nil {
		return nil, err
	}
```

pkg/registry/registry.go
```diff
@@ -110,147 +111,83 @@ func (r *registry) LastLayer(ctx context.Context, entry string) (v1.Layer, error
...
-	targz, err := layer.Compressed()
+	obj, err := r.getHeaderFromLayer(layer, "etc/driver-toolkit-release.json")
 	if err != nil {
-		return dtk, err
+		return nil, err
 	}
-	defer r.dclose(targz)
 
-	gr, err := gzip.NewReader(targz)
+	entry, _, err := unstructured.NestedString(obj.Object, "KERNEL_VERSION")
```
qbarrand, line 119:
```suggestion
	dtk.KernelFullVersion, _, err := unstructured.NestedString(obj.Object, "KERNEL_VERSION")
```

pkg/registry/registry.go
```diff
@@ -110,147 +111,83 @@ func (r *registry) LastLayer(ctx context.Context, entry string) (v1.Layer, error
...
-			}
-
-			tags, _, err := unstructured.NestedSlice(obj.Object, "spec", "tags")
-			if err != nil {
-				return "", "", err
+	for _, tag := range tags {
+		if tag.(map[string]interface{})["name"] == "driver-toolkit" {
+			from, ok := tag.(map[string]interface{})["from"]
+			if !ok {
+				return "", fmt.Errorf("invalid image reference format for driver toolkit entry, from tag")
```
qbarrand, line 154:
```suggestion
				return "", errors.New("invalid image reference format for driver toolkit entry: missing `from` tag")
```

pkg/registry/registry.go
```diff
@@ -110,147 +111,83 @@ func (r *registry) LastLayer(ctx context.Context, entry string) (v1.Layer, error
...
 			}
-
-			for _, tag := range tags {
-				if tag.(map[string]interface{})["name"] == "driver-toolkit" {
-					from := tag.(map[string]interface{})["from"]
-					imageURL = from.(map[string]interface{})["name"].(string)
-				}
+			imageURL, ok := from.(map[string]interface{})["name"].(string)
+			if !ok {
+				return "", fmt.Errorf("invalid image reference format for driver toolkit entry, name tag")
```
qbarrand, line 158:
```suggestion
				return "", errors.New("invalid image reference format for driver toolkit entry: missing `name` tag")
```

pkg/registry/registry.go
```diff
@@ -110,147 +111,83 @@ func (r *registry) LastLayer(ctx context.Context, entry string) (v1.Layer, error
...
 			}
-
+			return imageURL, nil
 		}
+	}
+	return "", fmt.Errorf("failed to find driver-toolkit in the release-manifests/image-references")
```
qbarrand, line 163:
```suggestion
	return "", errors.New("failed to find driver-toolkit in the release-manifests/image-references")
```

pkg/registry/registry.go
pacevedom, line 119: Should we check if its found? It is supposed to be there, but what if it doesnt?

Same applies for the next two.

pkg/registry/registry.go
```diff
@@ -110,147 +111,83 @@ func (r *registry) LastLayer(ctx context.Context, entry string) (v1.Layer, error
...
+func (r *registry) ReleaseManifests(layer v1.Layer) (string, error) {
+	obj, err := r.getHeaderFromLayer(layer, "release-manifests/image-references")
 	if err != nil {
-		return "", "", err
+		return "", err
 	}
-	defer r.dclose(targz)
 
-	gr, err := gzip.NewReader(targz)
+	tags, _, err := unstructured.NestedSlice(obj.Object, "spec", "tags")
```
pacevedom, line 146: Same comment.

pkg/registry/registry.go
```diff
@@ -110,147 +111,83 @@ func (r *registry) LastLayer(ctx context.Context, entry string) (v1.Layer, error
...
-			obj := unstructured.Unstructured{}
+	tags, _, err := unstructured.NestedSlice(obj.Object, "spec", "tags")
+	if err != nil {
+		return "", err
+	}
 
-			if err = json.Unmarshal(buff, &obj.Object); err != nil {
-				return "", "", err
+	for _, tag := range tags {
+		if tag.(map[string]interface{})["name"] == "machine-os-content" {
```
qbarrand, line 178: Do you think something like [tidwall](https://github.com/tidwall/gjson) could be helpful to search in this JSON?

pkg/registry/registry.go
```diff
@@ -110,147 +111,83 @@ func (r *registry) LastLayer(ctx context.Context, entry string) (v1.Layer, error
...
-				return "", "", err
-			}
+func (r *registry) ReleaseImageMachineOSConfig(layer v1.Layer) (string, error) {
+	obj, err := r.getHeaderFromLayer(layer, "release-manifests/image-references")
+	if err != nil {
+		return "", err
+	}
 
-			obj := unstructured.Unstructured{}
+	tags, _, err := unstructured.NestedSlice(obj.Object, "spec", "tags")
```
pacevedom, line 172: Same comment.

pkg/registry/registry.go
```diff
@@ -110,147 +111,83 @@ func (r *registry) LastLayer(ctx context.Context, entry string) (v1.Layer, error
...
+	return "", fmt.Errorf("failed to find driver-toolkit in the release-manifests/image-references")
+}
 
-		if header.Name == "release-manifests/release-metadata" {
-
-			buff, err := io.ReadAll(tr)
-			if err != nil {
-				return "", "", err
-			}
+func (r *registry) ReleaseImageMachineOSConfig(layer v1.Layer) (string, error) {
```
qbarrand, line 166: Could these methods be tested without too much hassle?

pkg/registry/registry.go
```diff
@@ -307,6 +244,49 @@ func (r *registry) GetLayerByDigest(repo string, digest string, auth []crane.Opt
 	return crane.PullLayer(repo+"@"+digest, auth...)
 }
 
+func (r *registry) getHeaderFromLayer(layer v1.Layer, headerName string) (*unstructured.Unstructured, error) {
```
pacevedom, line 247: Why header instead of file?

pkg/registry/registry.go
yevgeny-shnaidman: problematic, we need a prebuilt specific image and local repo for unit-test...

pkg/registry/registry.go
yevgeny-shnaidman: This is a name used by the tar package,  so i thought to keep it here

pkg/registry/registry.go
yevgeny-shnaidman: it is a nice package, but its error handling is problematic, i would prefer to stay with what we have right now. Since this is not the only place we are parsing json, we can open a Jira to investigate other packages, and then , if we find one, to replace project -wise

pkg/registry/registry.go
yevgeny-shnaidman: fixed

pkg/registry/registry.go
yevgeny-shnaidman: fixed

pkg/registry/registry.go
yevgeny-shnaidman: fixed

pkg/registry/registry.go
yevgeny-shnaidman: fixed

pkg/registry/registry.go
yevgeny-shnaidman: fixed

pkg/registry/registry.go
yevgeny-shnaidman: fixed

pkg/registry/registry.go
yevgeny-shnaidman: fixed

pkg/registry/registry.go
yevgeny-shnaidman: fixed

pkg/registry/registry.go
qbarrand: Yes, I think that makes sense. This is not the only place where we search for data in unstructured JSON indeed.

pkg/registry/registry.go
qbarrand: Is it difficult to build a dummy `v1.Layer`?

yevgeny-shnaidman: it means building v1.Layer and all its contents, so that gzip and tar package will work correctly

pkg/registry/registry.go
qbarrand: https://issues.redhat.com/browse/MGMT-9699

PR #146, opened 2022-03-21
Bug 2066359: SRO appends "arm64" instead of "aarch64" to the kernel name and it doesn't match the DTK

pacevedom:


pacevedom:
This issue is fixed in 4.11 and further, will handle BZ manually.

bthurber (approved):

PR #147, opened 2022-03-22
MGMT-9705: add cluster pkg API for preflight support

yevgeny-shnaidman:
This PR include the following changes:
1) new API IsClusterInUpgrade - check if the cluster is currently in the upgrade
2) new API NextUpgradeVersion - returns the OCP image that will be used for upgrade
3) remove validation if various object are present in the cluster, since in OCP cluster
   they should be present, and enability to access them should result in clear error

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -9,9 +9,7 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
-	configv1 "github.com/openshift/api/config/v1"
 	imagev1 "github.com/openshift/api/image/v1"
-	machinev1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	"github.com/openshift/special-resource-operator/pkg/clients"
 	"github.com/openshift/special-resource-operator/pkg/utils"
 
@@ -30,6 +28,8 @@ type Cluster interface {
 	OSImageURL(context.Context) (string, error)
 	OperatingSystem(*corev1.NodeList) (string, string, string, error)
 	GetDTKImages(context.Context) ([]string, error)
+	NextUpgradeVersion(context.Context) (string, error)
+	IsClusterInUpgrade(context.Context) (bool, error)
 }
 
 func NewCluster(clients clients.ClientsInterface) Cluster {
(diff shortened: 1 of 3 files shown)


pkg/cluster/cluster.go
```diff
@@ -163,15 +146,43 @@ func (c *cluster) OperatingSystem(nodeList *corev1.NodeList) (string, string, st
...
-	if !clusterVersionAvailable {
-		c.log.Info("Warning: ClusterVersion API resource not available. Can be ignored on vanilla k8s.")
+
+	desired := version.Status.Desired.Image
+	completedVersions := []configv1.UpdateHistory{}
+	for _, condition := range version.Status.History {
+		if condition.State != "Completed" {
+			continue
+		}
+		completedVersions = append(completedVersions, condition)
```
qbarrand, line 174:
```suggestion
		if condition.State == "Completed" {
			completedVersions = append(completedVersions, condition)
		}
```

pkg/cluster/cluster.go
```diff
@@ -163,15 +146,43 @@ func (c *cluster) OperatingSystem(nodeList *corev1.NodeList) (string, string, st
...
+	if len(completedVersions) == 0 {
+		// probably cluster is still in the final stages of installation
 		return false, nil
 	}
-	return true, nil
+
+	sort.Slice(completedVersions, func(i, j int) bool {
+		return completedVersions[i].CompletionTime.Time.After(completedVersions[j].CompletionTime.Time)
+	})
+	if completedVersions[0].Image != desired {
```
qbarrand, line 184: Can we avoid sorting here? If we want to take the latest completion time, then I think only one pass keeping track of the most recent version should be enough.

pkg/cluster/cluster.go
yevgeny-shnaidman: fixed

pkg/cluster/cluster.go
yevgeny-shnaidman: fixed

pkg/cluster/cluster.go
```diff
@@ -9,9 +9,8 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
-	configv1 "github.com/openshift/api/config/v1"
+	//configv1 "github.com/openshift/api/config/v1"
```
qbarrand, line 12: Could we remove this?

yevgeny-shnaidman: fixed

PR #149, opened 2022-03-25
MGMT-9662 Use Context's logger

pmtk:
Removes dedicated loggers from most of a packages.
Instead uses logger that is stored inside a Context instead.
Provides helper functions (based on controller-runtime's) for extracting and storing logger withing Context: in case of missing logger inside Context a just-in-case logger will be used: "ctx-missing-logger".


controllers/workitem.go
```diff
@@ -1,17 +1,20 @@
...
+	"github.com/go-logr/logr"
 	"helm.sh/helm/v3/pkg/chart"
 )
 
 // WorkItem stores values required for current reconciliation
 type WorkItem struct {
-	// Log is a logger dedicated for specific SpecialResource constructed with its NamespacedName.
-	Log logr.Logger
+	// Ctx is a context for reconciling specific SpecialResource. It stores a logger.
+	Ctx context.Context
```
qbarrand, line 17: Contexts should not be stored in structs: https://go.dev/blog/context-and-structs
Is there a reason why you decided not to use the `Log` field?

pmtk: This change is based on https://github.com/openshift/special-resource-operator/pull/125#discussion_r820641725.
I'm on a fence with this change, we can keep it as poc :)

Logger is packaged inside Ctx, so when Ctx is passed around to various modules (`pkg/`), the modules will use the same logger which name is SpecialResource name. In concurrent reconciliations that would help discerning what action is for which SpecialResource. 
Adding `logger` to every method of the interfaces in the `pkg/` seems redundant and less elegant, but on the other hand I don't like the implicit contract of Ctx storing logger.

pkg/utils/ctxlog.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"context"
+
+	"github.com/go-logr/logr"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+)
+
+var missingLogger = zap.New().WithName("ctx-missing-logger")
+
+func LoggerFromContext(ctx context.Context) logr.Logger {
```
qbarrand, line 12: Could [`LogFromContext`](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.11.2/pkg/log#FromContext) replace this?

internal/controllers/state/clusteroperator_test.go
```diff
@@ -12,10 +12,14 @@ import (
 	srov1beta1 "github.com/openshift/special-resource-operator/api/v1beta1"
 	"github.com/openshift/special-resource-operator/internal/controllers/state"
 	"github.com/openshift/special-resource-operator/pkg/clients"
+	"github.com/openshift/special-resource-operator/pkg/utils"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
 
+var ctx = utils.ContextWithLogger(context.TODO(), zap.New().WithName("test"))
```
qbarrand, line 21: Is this needed?

pkg/poll/poll.go
```diff
@@ -133,12 +129,12 @@ func (p *pollActions) ForResource(ctx context.Context, obj *unstructured.Unstruc
 	// Wait for general availability, Pods Complete, Running
 	// DaemonSet NumberUnavailable == 0, etc
 	if wait, ok := p.waitFor[obj.GetKind()]; ok {
-		p.log.Info("ForResource", "Kind", obj.GetKind())
+		utils.LoggerFromContext(ctx).Info("Waiting for resource", "kind", obj.GetKind(), "namespace", obj.GetNamespace(), "name", obj.GetName())
```
qbarrand, line 132: Aren't we receiving this from the reconciler? If so, name and namespace should already be populated?

pkg/watcher/watcher.go
```diff
@@ -71,21 +72,22 @@ func getAPIVersion(gvk schema.GroupVersionKind) string {
...
 type Watcher interface {
 	// ReconcileWatches is a general function that updates list of watched resources based on given SpecialResourceModule
-	ReconcileWatches(srov1beta1.SpecialResourceModule) error
+	ReconcileWatches(context.Context, srov1beta1.SpecialResourceModule) error
 }
 
 func New(ctrl controller.Controller) Watcher {
 	return &watcher{
-		log:               zap.New(zap.UseDevMode(true)).WithName(utils.Print("watcher", utils.Blue)),
+		mapperLog:         zap.New().WithName("watcher"),
```
qbarrand, line 80: Could we pass this from the constructor?

pkg/utils/ctxlog.go
pmtk: It's actually almost the same - difference is that if the context doesn't contain a logger they'll provide a null one and here it's providing a logger with 'ctx-missing-logger' name which hopefully would help us if this ever happens

internal/controllers/state/clusteroperator_test.go
pmtk: This is so the tests proceed to produce logs as it is now. We could drop it and use `context.TODO()` or `context.Background()` which would result in using `ctx-missing-logger`

pkg/utils/ctxlog.go
yevgeny-shnaidman: Also, it would be better not to have a global variable in the package. Maybe something like:
logger := logr.FromContext(ctx); 
if logger != nil {
    return logger
}
return  zap.New().WithName("ctx-missing-logger")

pkg/watcher/watcher.go
pmtk: Used `ctrl.GetLogger()`

pkg/preflight/preflight_test.go
```diff
@@ -1,7 +1,7 @@
 package preflight
 
 import (
-	"context"
+	context "context"
```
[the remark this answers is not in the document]
pmtk: Whoops! changed

pkg/filter/filter.go
```diff
@@ -28,7 +28,7 @@ type Filter interface {
 
 func NewFilter(kind, ownedLabel string, lifecycle lifecycle.Lifecycle, storage storage.Storage, kernelData kernel.KernelData) Filter {
 	return &filter{
-		log:        zap.New(zap.UseDevMode(true)).WithName(utils.Print("filter", utils.Purple)),
+		log:        zap.New().WithName("filter"),
```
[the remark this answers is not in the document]
pmtk: Changed

controllers/specialresource.go
```diff
@@ -76,17 +75,18 @@ func (r *SpecialResourceReconciler) SpecialResourcesReconcile(ctx context.Contex
 		return reconcile.Result{}, err
 	}
 
-	log.Info("Resolving dependencies")
+	log.Info("Resolving dependencies", "amount", len(wi.SpecialResource.Spec.Dependencies))
```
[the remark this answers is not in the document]
pmtk: changed

pkg/utils/ctxlog.go
pmtk: Changed according to Yevgeny's suggestion

controllers/specialresourcemodule_controller.go
```diff
@@ -314,7 +312,7 @@ func (r *SpecialResourceModuleReconciler) getVersionInfoFromImage(ctx context.Co
 }
 
 func (r *SpecialResourceModuleReconciler) getOCPVersions(ctx context.Context, watchList []srov1beta1.SpecialResourceModuleWatch) (map[string]ocpVersionInfo, error) {
-	logVersion := r.Log.WithName(utils.Print("versions", utils.Purple))
+	logVersion := utils.LoggerFromContext(ctx).WithName("versions")
```
yevgeny-shnaidman, line 315: Why are we setting the Name in this function? We did not do it in any other

pkg/kernel/kernel.go
```diff
@@ -122,7 +116,6 @@ func (k *kernelData) IsObjectAffine(obj client.Object) bool {
 	annotations := obj.GetAnnotations()
 
 	if affine, found := annotations["specialresource.openshift.io/kernel-affine"]; found && affine == "true" {
-		k.log.Info("Object is Kernel Affine", "Object", obj.GetName())
```
yevgeny-shnaidman, line 125: why do we not need that log?

pkg/resource/resource.go
```diff
@@ -482,14 +478,14 @@ func (r *resource) sendNodesMetrics(ctx context.Context, obj *unstructured.Unstr
 		nodesStr := strings.Join(nodesNames, ",")
 		r.metricsClient.SetUsedNodes(crName, obj.GetName(), obj.GetNamespace(), kind, nodesStr)
 	} else {
-		r.log.Info("No assigned nodes for found for UsedNodes metric", "kind", kind, "name", obj.GetName(), "crName", crName)
+		utils.LoggerFromContext(ctx).Info("No assigned nodes for found for UsedNodes metric", "kind", kind, "name", obj.GetName(), "crName", crName)
```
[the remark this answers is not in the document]
pmtk: Changed

pkg/resource/resource.go
```diff
@@ -413,12 +409,12 @@ func (r *resource) createObjFromYAML(
 	// We are asuming that vendors provide pre compiled DriverContainers
 	// If err == nil, build a new container, if err != nil skip it
 	if err = r.rebuildDriverContainer(obj); err != nil {
-		r.log.Info("Skipping building driver-container", "Name", obj.GetName())
+		utils.LoggerFromContext(ctx).Info("Skipping building driver-container", "Name", obj.GetName())
```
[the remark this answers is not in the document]
pmtk: Changed

pkg/kernel/kernel.go
pmtk: Added this log back in places that invoke that function (otherwise it'd need a `log` param)

controllers/specialresourcemodule_controller.go
pmtk: Removed

pkg/utils/warn.go
```diff
@@ -4,18 +4,18 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
 
-var log = zap.New(zap.UseDevMode(true)).WithName(Print("warning", Brown))
+var log = zap.New().WithName("warning")
```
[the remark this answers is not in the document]
qbarrand: I am in favor of removing it.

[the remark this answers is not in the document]
pmtk: Removed it. One problematic place is `func (r *registry) dclose(c io.Closer)` in `registry.go`. Please take a look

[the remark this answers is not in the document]
qbarrand: We're not writing to files on which we call `dclose()`, so I would remove that method entirely and do `defer c.Close()` directly. WDYT?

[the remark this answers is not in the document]
yevgeny-shnaidman: i agree with @qbarrand , we only open for reading, if closing failed, it is not really an issue, we can ignore the error in that case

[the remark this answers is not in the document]
yevgeny-shnaidman: lets just add a comment why we allow ourselves to ignore the error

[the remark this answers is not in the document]
pmtk: Removed the function and added comments

PR #150, opened 2022-03-27
MGMT-9766: change Makefile manager target output

yevgeny-shnaidman:
Currently manager target in the Makefile produces manager executable, therefore changing the source code files does not trigger rebuilt.
This PR changes the output executable to be sro_manager

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,6 +86,7 @@ helm-plugins/cm-getter: helm-plugins/cm-getter/cm-getter
 .PHONY: helm-plugins
 helm-plugins: helm-plugins/cm-getter
 
+.PHONY: manager
 manager:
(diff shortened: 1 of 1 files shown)


yevgeny-shnaidman:
@assign @ybettan

ybettan:
I will suggest using `.PHONY` instead of changing the the binary name.

yevgeny-shnaidman:
> I will suggest using `.PHONY` instead of changing the the binary name.

fixed

ybettan:
/lgtm
/hold
Let merge this change u/s first.

ybettan:
Here is the u/s [PR](https://github.com/openshift-psap/special-resource-operator/pull/277)

PR #151, opened 2022-03-27
Add proxy environment variables to Dockerfile to avoid Internet dependency

omertuc:
See comment inside Dockerfile for more information

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,28 @@
 # Build the manager binary
 FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.17-openshift-4.11 as builder
 
+# This Dockerfile is sometimes built in environments (ART) where there's no
+# Internet connectivity. By messing up Internet connectivity across the board
+# for everyone, we make sure that developers won't make a mistake of making
+# this Dockerfile build dependent on Internet connectivity. This current
+# implementation of the connectivity blocking is not airtight, it only affects
+# HTTP traffic and only affects programs which actually respect these proxy
+# environment variables, but it was chosen because it's simple and alternatives
+# are hard to come up with.
+#
+# The goal is only to block Internet connectivity for the build process itself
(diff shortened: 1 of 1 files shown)


ybettan:
That looks AWESOME @omertuc.
Thanks for that PR.
I will test it tomorrow to see how it goes.

ybettan:
/cc @qbarrand 
This looks like a valid option for us as it is only a small diff from u/s.

Dockerfile
```diff
@@ -1,6 +1,28 @@
...
 FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.17-openshift-4.11 as builder
 
+# This Dockerfile is sometimes built in environments (ART) where there's no
+# Internet connectivity. By messing up Internet connectivity across the board
+# for everyone, we make sure that developers won't make a mistake of making
+# this Dockerfile build dependent on Internet connectivity. This current
+# implementation of the connectivity blocking is not airtight, it only affects
+# HTTP traffic and only affects programs which actually respect these proxy
+# environment variables, but it was chosen because it's simple and alternatives
+# are hard to come up with.
```
qbarrand, line 11: Is this standard across Edge projects?
We were thinking about having a GitHub action instead that would use something like `unshare` to simulate a disconnected environment (Linux namespaces are supported there). How about that?

omertuc: >Is this standard across Edge projects?

No, I'm not aware of any project that tried to have a job validating their Dockerfile works even without an Internet connection. 

>We were thinking about having a GitHub action

This PR is a solution I proposed to get it to work in Prow, if GitHub actions is something you want to start depending on just for this then it's probable that there are better solutions than what I did.

ybettan:
/hold
We may use our current CI to build SRO inside the cluster using `build` resource.

omertuc:
Okay closing this, re-open if you ever want to

PR #154, opened 2022-03-29
MGMT-9792: add an OC executable to the skipper container

yevgeny-shnaidman:


ybettan:
Why do we need it? Can you write it in the commit message and PR description?
Should we have it u/s as well? If yes then please let's start with the u/s PR

yevgeny-shnaidman:
> Why do we need it? Can you write it in the commit message and PR description? Should we have it u/s as well? If yes then please let's start with the u/s PR

we don't need it in the upstream, since there is no openshift support there. We need it since @qbarrand set the OC as a default client in the makefile operartions

PR #155, opened 2022-03-29
MGMT-9798: Add makefile target for local test ConfigMap deployment

yevgeny-shnaidman:
Since SRO test chart has been moved to use ConfigMap instead of file,
we need to add a target to the Makefile that allows to create load ConfigMsap
containing the chart data (simple-kmod, ping-pong)
into the test cluster of the user (probably, on some beaker machine)
This PR add a target that allows loading such CM into cluster, and provides
infrastructure for expanding to other future recipes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -117,6 +117,14 @@ deploy: manifests kustomize configure ## Deploy controller in the configured Kub
 	$(shell sleep 5)
 	$(CLUSTER_CLIENT) apply -k config/cr
 
+deploy-cm-simple-kmod:
+	$(CLUSTER_CLIENT) create namespace simple-kmod
+	./scripts/make-cm-recipe charts/example/simple-kmod-0.0.1 simple-kmod-chart simple-kmod
+
+deploy-cm-ping-pong:
+	$(CLUSTER_CLIENT) create namespace ping-pong
+	./scripts/make-cm-recipe charts/example/ping-pong-0.0.1 ping-pong-chart ping-pong
+
 # If the CRD is deleted before the CRs the CRD finalizer will hang forever
 # The specialresource finalizer will not execute either
 undeploy: kustomize ## Undeploy controller from the K8s cluster specified in ~/.kube/config.


scripts/make-local-cm
```diff
@@ -0,0 +1,42 @@
...
+
+TMP="$(mktemp -d)"
+readonly TMP
+
+function generate_simple_kmod() {
+	helm package charts/example/simple-kmod-0.0.1/ -d "${TMP}"
+	helm repo index "${TMP}"
+	echo "Using namespace simple-kmod"
+	echo "Using name simple-kmod-chart"
+	kubectl create namespace simple-kmod
```
pacevedom, line 27: what about using `CLUSTER_CLIENT` instead of plain kubectl?

pacevedom, line 27: What do you think about using a create+apply pattern?
`kubectl create namespace simple-kmod -o yaml --dry-run=client | kubectl apply -f -`

Makefile
```diff
@@ -117,6 +117,12 @@ deploy: manifests kustomize configure ## Deploy controller in the configured Kub
 	$(shell sleep 5)
 	$(CLUSTER_CLIENT) apply -k config/cr
 
+deploy-cm-%:
+	KUBECONFIG="$(KUBECONFIG)" ./scripts/make-local-cm generate_$(subst -,_,$*)
```
ybettan, line 121: I don't think we need to set the `KUBECONFIG` var here, it should work without it IMO.
Also I never used the `subst` command before, but isn't it going to apply the configmap into the default namespace? I think we should add the namespace parameter here.

ybettan, line 121: I don't think we need `KUBECONFIG` here. It should work without it.

scripts/make-local-cm
```diff
@@ -0,0 +1,42 @@
+#!/bin/sh
```
ybettan, line 1: This script looks very similar to `scripts/make-cm-recipe`.
Is there a reason to create an additional script that basically does the same?

scripts/make-local-cm
ybettan: What is the benefit of that?

Makefile
yevgeny-shnaidman: fixed

scripts/make-local-cm
yevgeny-shnaidman: for now i prefer not to mix them together, since in local script we set all the variable in the function, and in e2e they can be derived . I would prefer to push it as is, and later on, we might try to merge them.

scripts/make-local-cm
yevgeny-shnaidman: fixed

pacevedom: It does not fail if the resource already exists, and updates it. With create you get an error if the namespace or the configmap already exist and the script stops.

yevgeny-shnaidman: fixed

PR #156, opened 2022-03-30
MGMT-9731: add helmer package API for preflight preprocessing

yevgeny-shnaidman:
For the preflight code we need to preprocess the helm charts of a CR
to get all the resulting yamls, without actually loading them into the cluster.
This PR does the following
1) add an API that just does the helm preprocessing and returns a list of yaml
2) change Run API to use the shared code of the new API


qbarrand:
What kind of errors would such a dry run catch? Isn't the danger rather in the DriverContainer image (or the lack of it) for the new kernel?

yevgeny-shnaidman:
> What kind of errors would such a dry run catch? Isn't the danger rather in the DriverContainer image (or the lack of it) for the new kernel?

The purpose of the API is to process the the helm chart and to produce the list of yamls without loading them into cluster. Preflight will use those yaml to decide what to check and how to check

PR #157, opened 2022-03-30
Adding a doc section on how to write good commit messages.

ybettan:
This is important to allow us to easily understand why changes where
made without the need of figuring out the code.


ybettan:
/cc @yevgeny-shnaidman 
Related u/s [PR](https://github.com/openshift-psap/special-resource-operator/pull/279)

PR #158, opened 2022-03-30
MGMT-9763: add resource pkg API to support preflight validations

yevgeny-shnaidman:
The changes in this PR:
1) add API to get a list of Objects from helm YAMLs
2) rename Creator interface to ResourceAPI, and update the code accordingly

### DIFF
--- a/pkg/resource/resource.go
+++ b/pkg/resource/resource.go
@@ -36,83 +36,84 @@ var (
 
 //go:generate mockgen -source=resource.go -package=resource -destination=mock_resource_api.go
 
-type Creator interface {
+type ResourceAPI interface {
 	CreateFromYAML(context.Context, []byte, bool, v1.Object, string, string, map[string]string, string, string) error
+	GetObjectsFromYAML([]byte) (*unstructured.UnstructuredList, error)
 }
 
-type creator struct {
+type resource struct {
 	kubeClient    clients.ClientsInterface
 	lc            lifecycle.Lifecycle
 	log           logr.Logger
 	metricsClient metrics.Metrics
 	pollActions   poll.PollActions
-	kernelData    kernel.KernelData
+	kernelAPI     kernel.KernelData
 	proxyAPI      proxy.ProxyAPI
 	scheme        *runtime.Scheme
 	helper        resourcehelper.Helper
 }
 
-func NewCreator(
+func NewResourceAPI(
 	kubeClient clients.ClientsInterface,
 	metricsClient metrics.Metrics,
 	pollActions poll.PollActions,
-	kernelData kernel.KernelData,
+	kernelAPI kernel.KernelData,
 	scheme *runtime.Scheme,
 	lc lifecycle.Lifecycle,
 	proxyAPI proxy.ProxyAPI,
 	resHelper resourcehelper.Helper,
-) Creator {
-	return &creator{
+) ResourceAPI {
+	return &resource{
 		kubeClient:    kubeClient,
 		lc:            lc,
 		log:           zap.New(zap.UseDevMode(true)).WithName(utils.Print("resource", utils.Blue)),
 		metricsClient: metricsClient,
 		pollActions:   pollActions,
-		kernelData:    kernelData,
+		kernelAPI:     kernelAPI,
 		scheme:        scheme,
 		proxyAPI:      proxyAPI,
 		helper:        resHelper,
 	}
 }
 
-func (c *creator) AfterCRUD(ctx context.Context, obj *unstructured.Unstructured, namespace string) error {
+func (r *resource) AfterCRUD(ctx context.Context, obj *unstructured.Unstructured, namespace string) error {
 
-	logger := c.log.WithValues("Kind", obj.GetObjectKind(), "namespace", obj.GetNamespace(), "name", obj.GetName())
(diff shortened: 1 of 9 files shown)


pkg/resource/resource.go
```diff
@@ -159,34 +160,58 @@ func (c *creator) CreateFromYAML(
...
 }
 
+func (r *resource) GetObjectsFromYAML(ctx context.Context, yamlFile []byte) (*unstructured.UnstructuredList, error) {
+	scanner := yamlutil.NewYAMLScanner(yamlFile)
+	objList := &unstructured.UnstructuredList{
+		Items: []unstructured.Unstructured{},
+	}
+	for scanner.Scan() {
+		yamlSpec := scanner.Bytes()
+		obj := &unstructured.Unstructured{
```
pmtk, line 170: Could creating this object on stack (without `&`) have some benefits in relieving GC? Not that this matters much with amount of objects ;)

pkg/resource/resource.go
```diff
@@ -159,34 +160,58 @@ func (c *creator) CreateFromYAML(
...
+			Object: map[string]interface{}{},
+		}
+
+		jsonSpec, err := yaml.YAMLToJSON(yamlSpec)
+		if err != nil {
+			return nil, fmt.Errorf("Could not convert yaml file to json: %s: error %w", string(yamlSpec), err)
```
pacevedom, line 176: Dumping the whole yaml might be hard to read as part of the error message. WDYT?

pkg/resource/resource.go
```diff
@@ -159,34 +160,58 @@ func (c *creator) CreateFromYAML(
...
+		}
+
+		if err = obj.UnmarshalJSON(jsonSpec); err != nil {
+			return nil, fmt.Errorf("cannot unmarshall json spec, check your manifest: %s: %w", jsonSpec, err)
```
pacevedom, line 180: Same here but with json.

pkg/resource/resource.go
yevgeny-shnaidman: We almost never see those errors (yaml mrshall/unmarshal, to json), so in that rare occurrence that it does happen, it worth to have all that data

pkg/resource/resource.go
yevgeny-shnaidman: it does not matter, but it is a good catch, does not make sense to create it as a pointer, and then to derefernce it

yevgeny-shnaidman: fixed

pkg/resource/resource.go
qbarrand:
```suggestion
			return nil, fmt.Errorf("could not convert YAML file to JSON %s: error %w", string(yamlSpec), err)
```

pkg/resource/resource.go
qbarrand:
```suggestion
			return nil, fmt.Errorf("cannot unmarshal JSON spec, check your manifest: %s: %w", jsonSpec, err)
```

pkg/resource/resource.go
```diff
@@ -159,34 +160,58 @@ func (c *creator) CreateFromYAML(
 	return nil
 }
 
+func (r *resource) GetObjectsFromYAML(ctx context.Context, yamlFile []byte) (*unstructured.UnstructuredList, error) {
```
qbarrand, line 163: I don't see `ctx` being used in the function.
```suggestion
func (r *resource) GetObjectsFromYAML(yamlFile []byte) (*unstructured.UnstructuredList, error) {
```

pkg/resource/resource_test.go
```diff
@@ -273,7 +273,75 @@ spec:
...
+`)
+	It("good flow", func() {
+		r := NewResourceAPI(kubeClient, metricsClient, pollActions, kernelData, scheme, mockLifecycle, proxyAPI, helper)
+		resultList, err := r.GetObjectsFromYAML(context.TODO(), yamlSpec)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(len(resultList.Items)).To(Equal(2))
+		for _, item := range resultList.Items {
+			Expect(item.GetName()).To(BeElementOf([]string{"dsName", "podName"}))
+			Expect(item.GetKind()).To(BeElementOf([]string{"DaemonSet", "Pod"}))
+		}
```
qbarrand, line 340: I think the order should be deterministic here?
```suggestion
		Expect(resultList.Items[0].GetName()).To(Equal("podName"))
		Expect(resultList.Items[0].GetKind()).To(Equal("Pod"))
		Expect(resultList.Items[1].GetName()).To(Equal("dsName"))
		Expect(resultList.Items[1].GetKind()).To(Equal("DaemonSet"))
```

qbarrand (commented):
LGTM, just the usual pickiness.

pkg/resource/resource.go
yevgeny-shnaidman: fixed

pkg/resource/resource.go
yevgeny-shnaidman: fixed

pkg/resource/resource.go
yevgeny-shnaidman: fixed

pkg/resource/resource_test.go
yevgeny-shnaidman: fixed

PR #159, opened 2022-04-01
MGMT-9825: create preflight package, responsible for preflight check …

yevgeny-shnaidman:
…for single CR

This PR includes the following:
1) preflight package that executes preflight check for single PR
   a) compile helm chart
   b) get list of all yamls from the compiled chart
   c) go over the list of yamls and validate daemonset image, in case buildconfig is not present
2) add Init API to runtime, used by preflight
3) add mock file for helmer package


pkg/preflight/preflight.go
```diff
@@ -0,0 +1,295 @@
...
+	metricsAPI     metrics.Metrics
+	runtimeAPI     runtime.RuntimeAPI
+	kernelAPI      kernel.KernelData
+}
+
+func (p *preflight) PreflightUpgradeCheck(ctx context.Context,
+	sr *srov1beta1.SpecialResource,
+	runInfo *runtime.RuntimeInformation) error {
+
+	p.log.Info("start preflight for cr", "name", sr.Name)
```
pmtk, line 73: nit: other logs start with uppercase letter

pkg/preflight/preflight.go
```diff
@@ -0,0 +1,295 @@
...
+	k8sv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	k8sruntime "k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+)
+
+//go:generate mockgen -source=preflight.go -package=preflight -destination=mock_preflight_api.go
+
+type PreflightAPI interface {
+	PreflightUpgradeCheck(ctx context.Context,
```
pmtk, line 30: I don't see it being used anywhere. This will be integrated in the future?

ybettan:
Typo in the commit message:
preflight package that executes preflight check for single **PR** --> preflight package that executes preflight check for single **CR**

pkg/preflight/preflight.go
```diff
@@ -0,0 +1,295 @@
...
+func (p *preflight) PreflightUpgradeCheck(ctx context.Context,
+	sr *srov1beta1.SpecialResource,
+	runInfo *runtime.RuntimeInformation) error {
+
+	p.log.Info("start preflight for cr", "name", sr.Name)
+	sr.DeepCopyInto(&runInfo.SpecialResource)
+
+	chart, err := p.helmerAPI.Load(sr.Spec.Chart)
+	if err != nil {
+		p.log.Info("Failed to load helm chart for CR", "name", sr.Name)
```
ybettan, line 78:
```suggestion
		p.log.WithError(err).Error("Failed to load helm chart for CR", "name", sr.Name)
```

pkg/preflight/preflight.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+
+	yamlsList, err := p.processFullChartTemplates(ctx, chart, sr.Spec.Set, runInfo, sr.Namespace, runInfo.KernelFullVersion)
+	if err != nil {
+		p.log.Info("Failed to process full chart during preflight check")
+		return err
+	}
+	verified, err := p.handleYAMLsCheck(ctx, yamlsList, runInfo.KernelFullVersion)
+	if err != nil {
+		p.log.Info("Failed to verify the chart during preflight check", "err", err)
```
ybettan, line 89:
```suggestion
		p.log.Error(err, "Failed to verify the chart during preflight check", "err", err)
```

pkg/preflight/preflight.go
```diff
@@ -0,0 +1,295 @@
...
+}
+
+func (p *preflight) processFullChartTemplates(ctx context.Context,
+	chart *helmchart.Chart,
+	values unstructured.Unstructured,
+	runInfo *runtime.RuntimeInformation,
+	namespace string,
+	upgradeKernelVersion string) (string, error) {
+	var err error
+	// [TODO] - check that templates are not messed up during helm processing
```
ybettan, line 105: Are we doing it in another PR?

pkg/preflight/preflight.go
```diff
@@ -0,0 +1,295 @@
...
+	nodeVersion := runInfo.ClusterUpgradeInfo[upgradeKernelVersion]
+
+	runInfo.ClusterVersionMajorMinor = nodeVersion.ClusterVersion
+	runInfo.OperatingSystemDecimal = nodeVersion.OSVersion
+	runInfo.OperatingSystemMajorMinor = nodeVersion.OSMajorMinor
+	runInfo.OperatingSystemMajor = nodeVersion.OSMajor
+	runInfo.DriverToolkitImage = nodeVersion.DriverToolkit.ImageURL
+
+	fullChart.Values, err = chartutil.CoalesceValues(&fullChart, values.Object)
+	if err != nil {
```
ybettan, line 121: Can we log the error?

pkg/preflight/preflight.go
```diff
@@ -0,0 +1,295 @@
...
+		return "", err
+	}
+	rinfo, err := k8sruntime.DefaultUnstructuredConverter.ToUnstructured(&runInfo)
+	if err != nil {
```
ybettan, line 125: Log the error?

pkg/preflight/preflight.go
```diff
@@ -0,0 +1,295 @@
...
+		return "", err
+	}
+
+	fullChart.Values, err = chartutil.CoalesceValues(&fullChart, rinfo)
+	if err != nil {
```
ybettan, line 130: Log the error?

pkg/preflight/preflight.go
```diff
@@ -0,0 +1,295 @@
...
+		return "", err
+	}
+
+	return p.helmerAPI.GetHelmOutput(ctx, fullChart, fullChart.Values, namespace)
+}
+
+//[TODO] - handle multiple daemonsets and buildconfigs
```
ybettan, line 137: In a different PR?

pkg/preflight/preflight.go
```diff
@@ -0,0 +1,295 @@
...
+func (p *preflight) handleYAMLsCheck(ctx context.Context, yamlsList string, upgradeKernelVersion string) (bool, error) {
+	objList, err := p.resourceAPI.GetObjectsFromYAML([]byte(yamlsList))
+	if err != nil {
```
ybettan, line 140: Log?

pkg/preflight/preflight.go
```diff
@@ -0,0 +1,295 @@
...
+
+		p.log.Info("dtk info present in layer", "layerIndex", i)
+		if dtk.KernelFullVersion != upgradeKernelVersion {
+			p.log.Info("DTK kernel version differs from the upgrade node version", "dtkVersion", dtk.KernelFullVersion, "upgradeVersion", upgradeKernelVersion)
+			return false, nil
+		}
+		return true, nil
+	}
+
+	p.log.Info("DTK info not present on any layer of the image, not good", "image", image)
```
pmtk, line 203: This means that slim driver containers image (for example two stage builds with only *.ko and kmod in final image) are considered unverified but will still work?

pkg/preflight/preflight.go
```diff
@@ -0,0 +1,295 @@
...
+	for _, obj := range objList.Items {
+		kind := obj.GetKind()
+		if kind == "BuildConfig" {
+			// no more need to check daemons set, build config is present
+			p.log.Info("preflight: buildconfig related to daemonset, skipping image verification")
+			break
+		}
+		if kind == "DaemonSet" {
+			verified, err = p.daemonSetPreflightCheck(ctx, &obj, upgradeKernelVersion)
+			if err != nil {
```
ybettan, line 153: Log?

pkg/preflight/preflight.go
```diff
@@ -0,0 +1,295 @@
...
+		}
+	}
+	return verified, nil
+}
+
+func (p *preflight) daemonSetPreflightCheck(ctx context.Context, obj *unstructured.Unstructured, upgradeKernelVersion string) (bool, error) {
+	var ds k8sv1.DaemonSet
+	err := k8sruntime.DefaultUnstructuredConverter.FromUnstructured(obj.Object, &ds)
+	if err != nil {
+		p.log.Info("failed to convert YAML daemonset into sturct daemonset", "error", err)
```
ybettan, line 166:
```suggestion
		p.log.Error(er, "failed to convert YAML daemonset into sturct daemonset", "error", err)
```

pkg/preflight/preflight.go
```diff
@@ -0,0 +1,295 @@
...
+		return false, err
+	}
+	if ds.Spec.Template.Spec.Containers == nil || len(ds.Spec.Template.Spec.Containers) == 0 {
+		p.log.Info("invalid daemonset, no container  data present")
```
ybettan, line 170:
```suggestion
		p.log.Error(nil, "invalid daemonset, no container data present")
```

pkg/preflight/preflight.go
```diff
@@ -0,0 +1,295 @@
...
+			return false, nil
+		}
+		return true, nil
+	}
+
+	p.log.Info("DTK info not present on any layer of the image, not good", "image", image)
+	return false, nil
+}
+
+func (p *preflight) PrepareRuntimeInfo(ctx context.Context, image string) (*runtime.RuntimeInformation, error) {
```
pmtk, line 207: Question: So this is a hypothetical runtime info after the upgrade?

pkg/preflight/preflight.go
```diff
@@ -0,0 +1,295 @@
...
+		p.log.Info("invalid daemonset, no container  data present")
+		return false, fmt.Errorf("invalid daemonset, no container  data present")
+	}
+	image := ds.Spec.Template.Spec.Containers[0].Image
+
+	p.log.Info("daemonset image for preflight validation", "image", image)
+
+	repo, digests, auth, err := p.registryAPI.GetLayersDigests(ctx, image)
+	if err != nil {
+		p.log.Info("Failed to get layers digests for image", "image", image, "err", err)
```
ybettan, line 179:
```suggestion
		p.log.Error(err, "Failed to get layers digests for image", "image", image)
```

pkg/preflight/preflight.go
```diff
@@ -0,0 +1,295 @@
...
+		return false, nil
+	}
+
+	for i := len(digests) - 1; i >= 0; i-- {
+		layer, err := p.registryAPI.GetLayerByDigest(repo, digests[i], auth)
+		if err != nil {
+			p.log.Info("Failed to extract/access image", "image", image, "err", err)
```
ybettan, line 186:
```suggestion
			p.log.Error(err, "Failed to extract/access image", "image", image)
```

pkg/preflight/preflight.go
```diff
@@ -0,0 +1,295 @@
...
+			return false, nil
+		}
+		dtk, err := p.registryAPI.ExtractToolkitRelease(layer)
+		if err != nil {
+			p.log.Info("dtk info not present", "image", image, "layerIndex", i)
```
ybettan, line 191:
```suggestion
			p.log.Error(err, "dtk info not present", "image", image, "layerIndex", i)
```

pkg/preflight/preflight.go
```diff
@@ -0,0 +1,295 @@
...
+
+	p.log.Info("daemonset image for preflight validation", "image", image)
+
+	repo, digests, auth, err := p.registryAPI.GetLayersDigests(ctx, image)
+	if err != nil {
+		p.log.Info("Failed to get layers digests for image", "image", image, "err", err)
+		return false, nil
+	}
+
+	for i := len(digests) - 1; i >= 0; i-- {
```
ybettan, line 183: Not critical at all but is there a reason we go backward?

pkg/preflight/preflight.go
```diff
@@ -0,0 +1,295 @@
...
+		}
+		dtk, err := p.registryAPI.ExtractToolkitRelease(layer)
+		if err != nil {
+			p.log.Info("dtk info not present", "image", image, "layerIndex", i)
+			continue
+		}
+
+		p.log.Info("dtk info present in layer", "layerIndex", i)
+		if dtk.KernelFullVersion != upgradeKernelVersion {
+			p.log.Info("DTK kernel version differs from the upgrade node version", "dtkVersion", dtk.KernelFullVersion, "upgradeVersion", upgradeKernelVersion)
```
ybettan, line 197: Not sure if we should log as an error here as well.

pkg/preflight/preflight.go
```diff
@@ -0,0 +1,295 @@
...
+		return "", err
+	}
+
+	return dtk.KernelFullVersion, err
+}
+
+func (p *preflight) alertUserIfNeeded(verified bool, crName string) {
+	if verified {
+		p.log.Info("preflight check validation success, disabling alert", "crName", crName)
+		p.metricsAPI.SetUpgradeAlert(crName, 0)
```
ybettan, line 290: Isn't it the default value? do we really need to set it to zero manually?

pkg/preflight/preflight.go
yevgeny-shnaidman: Not sure yet. Need to understand if we really need, especially since this issue won't exists in V2

pkg/runtime/runtime.go
```diff
@@ -55,6 +55,7 @@ type RuntimeInformation struct {
 type RuntimeAPI interface {
 	GetRuntimeInformation(ctx context.Context, sr *srov1beta1.SpecialResource) (*RuntimeInformation, error)
 	LogRuntimeInformation(info *RuntimeInformation)
+	InitRuntimeInfo() *RuntimeInformation
```
ybettan, line 58: Why is that change needed?
Even if we do, I don't see any use of `InitRuntimeInfo` outside of this package - does it have to be an interface method?

pkg/preflight/preflight.go
yevgeny-shnaidman: current code work, this might be an optimization, but it won't be relevant for V2, so i'll just remove the TODO

pkg/preflight/preflight.go
yevgeny-shnaidman: Still thinking about it. Not sure if there is a recipe with multiple DS and BuildConfig in a chart, and it will surely not be an issue in the V2

pkg/preflight/preflight.go
```diff
@@ -0,0 +1,295 @@
...
+
+	runInfo.OSImageURL, err = p.clusterAPI.OSImageURL(ctx)
+	if err != nil {
+		p.log.Error(err, "failed to os image url for preflight")
+		return nil, err
+	}
+
+	// no need for real proxy configuration, since preflight check do not
+	// setup any object, just verifiying images
+	runInfo.Proxy = proxy.Configuration{}
```
qbarrand, line 243: Do we need to set it then? It's not a pointer, so `p.runtimeAPI.InitRuntimeInfo` has set it to the zero value already.

pkg/preflight/preflight.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+
+	runInfo.ClusterVersion, runInfo.ClusterVersionMajorMinor, err = p.clusterAPI.Version(ctx)
+	if err != nil {
+		p.log.Error(err, "Failed to get cluster version info for preflight")
+		return nil, err
+	}
+	// for preflight we do not need a real push secret, since preflight validation is not
+	// doing any image uploading
+	runInfo.PushSecretName = "dummyPushSecret"
```
qbarrand, line 233: Why do we need to set it?

pkg/preflight/preflight.go
```diff
@@ -0,0 +1,295 @@
...
+	runInfo.Proxy = proxy.Configuration{}
+
+	return runInfo, nil
+}
+
+func (p *preflight) getOSData(ctx context.Context, image string) (string, string, string, error) {
+	layer, err := p.registryAPI.LastLayer(ctx, image)
+	if err != nil {
+		p.log.Error(err, "failed to get last layer of image", "image", image)
+		return "", "", "", err
```
qbarrand, line 252: WDYT about returning the error messages instead of logging them?

pkg/preflight/preflight.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+	return utils.ParseOSInfo(machineOSConfig)
+}
+
+func (p *preflight) getKernelFullVersion(ctx context.Context, image string) (string, error) {
+	layer, err := p.registryAPI.LastLayer(ctx, image)
+	if err != nil {
+		p.log.Error(err, "failed to get last layer of image", "image", image)
+		return "", err
```
qbarrand, line 267: Same remark about logging vs returning the error messages.

pkg/preflight/preflight.go
ybettan: I prefer logging and returning the "virgin" error message with no wrappers around it.
Find it easier to track the root cause of an error.

When wrapping error, you may miss a few layers as it is wrapped in multiple places by multiple contributors but logging the error is easier to grep and you know the error is always the "root" error.

pkg/preflight/preflight.go
yevgeny-shnaidman: yes, this PR just introduces a package, that will be used by the preflight controller

pkg/preflight/preflight.go
yevgeny-shnaidman: fixed

pkg/preflight/preflight.go
yevgeny-shnaidman: fixed

pkg/preflight/preflight.go
yevgeny-shnaidman: fixed

pkg/preflight/preflight.go
yevgeny-shnaidman: fixed

pkg/preflight/preflight.go
yevgeny-shnaidman: fixed

pkg/preflight/preflight.go
```diff
@@ -0,0 +1,295 @@
...
+
+	chart, err := p.helmerAPI.Load(sr.Spec.Chart)
+	if err != nil {
+		p.log.Info("Failed to load helm chart for CR", "name", sr.Name)
+		return err
+	}
+
+	yamlsList, err := p.processFullChartTemplates(ctx, chart, sr.Spec.Set, runInfo, sr.Namespace, runInfo.KernelFullVersion)
+	if err != nil {
+		p.log.Info("Failed to process full chart during preflight check")
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

pkg/preflight/preflight.go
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

pkg/preflight/preflight.go
yevgeny-shnaidman: fixed

pkg/preflight/preflight.go
yevgeny-shnaidman: our requirement ( will be documented in the Release docs) is that if the customer uses slim images, he must still set the etc/driver-toolkit-release.json file into the image. This can be easily done if the customer uses DTK for building (one-liner).

pkg/preflight/preflight.go
yevgeny-shnaidman: Yes/ There are some minor fields that are not filled, since they don't really matter (like proxy, for example), because we are not actually loading the yamls into the cluster

pkg/preflight/preflight.go
yevgeny-shnaidman: fixed

pkg/preflight/preflight.go
yevgeny-shnaidman: fixed

pkg/preflight/preflight.go
yevgeny-shnaidman: This is not an error from the flow point of view, we don't know in which layer DTK info is present

pkg/preflight/preflight.go
ybettan: I was wondering the same thing... I can see the benefits of it but maybe it is redundant.

pkg/preflight/preflight.go
yevgeny-shnaidman: fixed

pkg/preflight/preflight.go
yevgeny-shnaidman: you are probably right, but i prefer to do it the same way as we do it in regular chart processing.  In any way, this code won't be needed in V2

pkg/preflight/preflight.go
yevgeny-shnaidman: Once the CR is fixed  by the user,  SRO will pass over it again, and i case it was verified, we want to turn off the alarm

pkg/runtime/runtime.go
yevgeny-shnaidman: It is used in the preflight package, to initialize the RuntimeInfo

pkg/preflight/preflight.go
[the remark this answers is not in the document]
yevgeny-shnaidman: I prefer to deal with the logging in a unified epic for logging that Chris Procter will introduce

pkg/preflight/preflight.go
```diff
@@ -0,0 +1,295 @@
...
+	chart *helmchart.Chart,
+	values unstructured.Unstructured,
+	runInfo *runtime.RuntimeInformation,
+	namespace string,
+	upgradeKernelVersion string) (string, error) {
+	var err error
+	// [TODO] - check that templates are not messed up during helm processing
+	fullChart := *chart
+	fullChart.Templates = []*helmchart.File{}
+	fullChart.Templates = append(fullChart.Templates, chart.Templates...)
```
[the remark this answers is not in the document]
yevgeny-shnaidman: i am not a fan of copy function, since it requires user to allocate the needed space for destination slice. Append takes care of that all by itself

pkg/preflight/preflight.go
```diff
@@ -0,0 +1,295 @@
...
+//[TODO] - handle multiple daemonsets and buildconfigs
+func (p *preflight) handleYAMLsCheck(ctx context.Context, yamlsList string, upgradeKernelVersion string) (bool, error) {
+	objList, err := p.resourceAPI.GetObjectsFromYAML([]byte(yamlsList))
+	if err != nil {
+		return false, err
+	}
+	verified := true
+	for _, obj := range objList.Items {
+		kind := obj.GetKind()
+		if kind == "BuildConfig" {
```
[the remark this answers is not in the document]
yevgeny-shnaidman: it can work but it won't be convenient: break inside switch will break only from the switch, and not from the upper loop. We can start using labels, but since it is only 2 ifs, i think staying with the ifs instead of switch should be ok

pkg/preflight/preflight.go
```diff
@@ -0,0 +1,295 @@
...
+}
+
+func (p *preflight) daemonSetPreflightCheck(ctx context.Context, obj *unstructured.Unstructured, upgradeKernelVersion string) (bool, error) {
+	var ds k8sv1.DaemonSet
+	err := k8sruntime.DefaultUnstructuredConverter.FromUnstructured(obj.Object, &ds)
+	if err != nil {
+		p.log.Info("failed to convert YAML daemonset into sturct daemonset", "error", err)
+		return false, err
+	}
+	if ds.Spec.Template.Spec.Containers == nil || len(ds.Spec.Template.Spec.Containers) == 0 {
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

pkg/preflight/preflight.go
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

pkg/preflight/preflight.go
```diff
@@ -0,0 +1,295 @@
...
+		p.log.Info("invalid daemonset, no container  data present")
+		return false, fmt.Errorf("invalid daemonset, no container  data present")
+	}
+	image := ds.Spec.Template.Spec.Containers[0].Image
+
+	p.log.Info("daemonset image for preflight validation", "image", image)
```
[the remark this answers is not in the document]
yevgeny-shnaidman: i would leave all the logger issues for Chris Procter epic

pkg/preflight/preflight.go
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

pkg/preflight/preflight.go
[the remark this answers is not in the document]
yevgeny-shnaidman: This is not an error, DTK info might be missing in a layer

pkg/preflight/preflight.go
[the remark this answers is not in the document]
yevgeny-shnaidman: as we discussed, customer should still provide etc/driver-toolkit-release.json in the multi-stage image

pkg/preflight/preflight.go
yevgeny-shnaidman: fixed

pkg/preflight/preflight.go
yevgeny-shnaidman: i think we should do both, especially the first time we encounter the error

pkg/preflight/preflight.go
yevgeny-shnaidman: i think we should do both, especially the first time we encounter the error

pkg/preflight/preflight.go
qbarrand: I think doing both would be:
```suggestion
		p.log.Error(err, "failed to get last layer of image", "image", image)
		return "", "", "", fmt.Errorf("failed to get last layer of image %s: %v", image, err)
)
```
WDYT?

pkg/preflight/preflight.go
ybettan: Are we keeping the TODO then?

PR #160, opened 2022-04-03
Adding CI artifacts.

ybettan:
We will deploy a podman container in the CI ephemeral cluster in order to
build SRO in a disconnected env.

This pod has to mount the pull-secret secret of the cluster, therefore, it
can't be initiated via cli as there is no option to mount a volume
on the pod's creation via cli and k8s doesn't supporting patching volumes
on running pods.

### DIFF
--- a/ci/podman.yaml
+++ b/ci/podman.yaml
@@ -1,0 +1,20 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: podman
+spec:
+  containers:
+  - name: podman
+    image: registry.redhat.io/rhel8/podman:latest
+    command:
+    - /usr/sbin/init
+    securityContext:
+      privileged: true
+    volumeMounts:
+    - name: pull-secret
+      mountPath: "/run/user/0/containers/"
+      readOnly: true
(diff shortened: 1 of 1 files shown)


ybettan:
/cc @qbarrand
/cc @yevgeny-shnaidman 

Once merged, I will be able to update the [CI's PR](https://github.com/openshift/release/pull/27410) to use the formal yaml.

/hold for CI to pass with remote yaml and make sure the flow works.

ci/podman.yaml
```diff
@@ -0,0 +1,20 @@
...
+metadata:
+  name: podman
+spec:
+  containers:
+  - name: podman
+    image: registry.redhat.io/rhel8/podman:latest
+    command:
+    - /usr/sbin/init
+    securityContext:
+      privileged: true
```
qbarrand, line 12: Why does it have to be privileged?

ybettan: Podman in podman requires privilege mode, otherwise you can't access the BE engine of building images AFAIK

PR #166, opened 2022-04-08
Use Go 1.18

qbarrand:
Upgrade to Go 1.18.
Use `go install` in `Makefile` to get dependences.
Upgrade `kustomize` to a version that works with `go install`.


qbarrand:
Closing as SRO is being sunset.

PR #168, opened 2022-04-10
MGMT-9241: adding preflight controller and reconcilaition loop

yevgeny-shnaidman:
This PR contains the following:
1) definition of the SpecialResourcePreflight CRD and it rbac configuration\
2) Preflight controller and reconcilation loop
3) changes to the preflight package, used by the Preflight reconcilation loop
4) additions to the statusupdater package

The flow of preflight reconcilaition loop:
1) get the preflight CR
2) add entry for each new  SRO CR to the preflight Status
   and set its status to Unknown
   (new SRO CR == never processed yet by current preflight CR)
3) for each SRO CR in the cluster - run preflight verification
   versus the image supplied by the Preflight CR, and update
   the verification status in the Preflight CR Status field
4) reschedule next reconcile loop in 1 minutes, unless all
   SRO CRs has been verified to True


controllers/specialresourcepreflight_controller.go
```diff
@@ -0,0 +1,214 @@
...
+		}).
+		Complete(r)
+}
+
+// Reconcile Reconiliation entry point
+func (r *SpecialResourcePreflightReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+
+	reconRes := ctrl.Result{}
+
+	log := r.Log.WithName(utils.Print("clusterpreflight", utils.Brown))
```
pmtk, line 66: Seems like we have two loggers here (that and `r.Log`). Is this a temporary until we have a logging strategy?

pkg/preflight/preflight.go
```diff
@@ -279,13 +273,3 @@ func (p *preflight) getKernelFullVersion(ctx context.Context, image string) (str
 
 	return dtk.KernelFullVersion, err
 }
-
-func (p *preflight) alertUserIfNeeded(verified bool, crName string) {
-	if verified {
-		p.log.Info("preflight check validation success, disabling alert", "crName", crName)
-		p.metricsAPI.SetUpgradeAlert(crName, 0)
```
pmtk, line 286: So we can remove `p.metricsAPI`? It's unused now

controllers/specialresourcepreflight_controller.go
```diff
@@ -0,0 +1,214 @@
...
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func (r *SpecialResourcePreflightReconciler) checkPreflightCompletion(ctx context.Context, name, namespace string) (bool, error) {
+	srp := srov1beta1.SpecialResourcePreflight{}
+	err := r.KubeClient.Get(ctx, types.NamespacedName{Namespace: namespace, Name: name}, &srp)
```
pmtk, line 191: Why get the object from API instead of using one that we already have (and was used to update Statuses)? Sanity/safety check?

api/v1beta1/specialresourcepreflight_types.go
```diff
@@ -0,0 +1,112 @@
...
+	// +kubebuilder:validation:Required
+	UpdateImage string `json:"updateImage"`
+}
+
+type SROStatus struct {
+	// name of SRO being checked
+	// +required
+	Name string `json:"name"`
+
+	// status of SRO verification: true (verified), false( verification failed), unknown
```
pmtk, line 53: Nit: I think the docstring should begin with uppercase. Also `false( verification` -> `false (verification`.
(Same comment for `StatusReason` below)

api/v1beta1/specialresourcepreflight_types.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+const (
+	VerificationTrue    string = "True"
+	VerificationFalse   string = "False"
+	VerificationError   string = "Error"
+	VerificationUnknown string = "Unknown"
+)
+
+// SpecialResourceSpec describes the desired state of the resource, such as the chart to be used and a selector
```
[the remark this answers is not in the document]
qbarrand: Or `and a node selector`?

api/v1beta1/specialresourcepreflight_types.go
```diff
@@ -0,0 +1,112 @@
...
+// on which nodes it should be installed.
+// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#spec-and-status
+// +kubebuilder:validation:Required
+type SpecialResourcePreflightSpec struct {
+	// Debug enables additional logging.
+	// +kubebuilder:validation:Optional
+	Debug bool `json:"debug"`
```
qbarrand, line 37: Not a fan of these fields being present in a CRD. This is rather a property of the operator, IMHO.

api/v1beta1/specialresourcepreflight_types.go
```diff
@@ -0,0 +1,112 @@
...
+
+	// Namespace describes in which namespace the chart will be installed.
+	// +kubebuilder:validation:Required
+	Namespace string `json:"namespace"`
+
+	// UpdateImage describe the OCP image that all SRO need to be checked against.
```
qbarrand, line 43: I don't understand the comment.

api/v1beta1/specialresourcepreflight_types.go
```diff
@@ -0,0 +1,112 @@
...
+	// +required
+	Name string `json:"name"`
+
+	// status of SRO verification: true (verified), false( verification failed), unknown
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum=True;False;Error;Unknown
+	VerificationStatus string `json:"verificationStatus"`
+
+	// reason contains a string describing the status source( if relevant).
```
qbarrand, line 59:
```suggestion
	// reason contains a string describing the status source (if relevant).
```

api/v1beta1/specialresourcepreflight_types.go
```diff
@@ -0,0 +1,112 @@
...
+
+// SpecialResourcePreflightStatus is the most recently observed status of the SpecialResourcePreflight.
+// It is populated by the system and is read-only.
+// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#spec-and-status
+type SpecialResourcePreflightStatus struct {
+	// SROStatuses contain observations about each SpecialResource's preflight upgradability check
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +optional
+	SROStatuses []SROStatus `json:"sroStatuses,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
```
qbarrand, line 80: I would drop the O in SROStatuses here as well.

api/v1beta1/specialresourcepreflight_types.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+
+// SpecialResourcePreflight initiates a preflight checks for all SpecialResources on the current Kuberentes cluster.
+// +kubebuilder:resource:path=specialresourcespreflight,scope=Cluster
+// +kubebuilder:resource:path=specialresourcespreflight,scope=Cluster,shortName=srp
+type SpecialResourcePreflight struct {
```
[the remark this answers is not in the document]
qbarrand: Or `PreflightCheck`.

controllers/specialresourcepreflight_controller.go
qbarrand: I think there's no need to keep a `Log` field in the controller struct. The manager already injects one in `ctx`, so this should work:
```suggestion
	log := logger := log.FromContext(ctx)
```

controllers/specialresourcepreflight_controller.go
```diff
@@ -0,0 +1,214 @@
...
+
+	log := r.Log.WithName(utils.Print("clusterpreflight", utils.Brown))
+	log.Info("ClusterPreflightController Request", "Name", req.Name, "Namespace", req.Namespace)
+
+	srp := srov1beta1.SpecialResourcePreflight{}
+	err := r.KubeClient.Get(ctx, types.NamespacedName{Namespace: req.Namespace, Name: req.Name}, &srp)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Info("SpecialResourcePreflight not found - probably deleted. Not reconciling", "name", req.Name, "namespace", req.Namespace)
+			return reconRes, nil
```
qbarrand, line 74: We could just skip reconciliations for deletions using a predicate when setting up the controller.

controllers/specialresourcepreflight_controller.go
```diff
@@ -0,0 +1,214 @@
...
+		} else {
+			log.Error(err, "Cluster preflight reconcile failed to find object", "name", req.Name, "namespace", req.Namespace)
```
qbarrand, line 76: If you use the context's logger, those fields are already filled for you.

controllers/specialresourcepreflight_controller.go
```diff
@@ -0,0 +1,214 @@
...
+	sroStatuses := r.getSROStatusesMap(srp)
+
+	runInfo, err := r.PreflightAPI.PrepareRuntimeInfo(ctx, srp.Spec.UpdateImage)
+	if err != nil {
+		r.Log.Info("Failed to get runtime info from image", "image", srp.Spec.UpdateImage)
+		return false, err
+	}
+
+	opts := []client.ListOption{}
+	err = r.KubeClient.List(ctx, specialresources, opts...)
```
qbarrand, line 113:
```suggestion
	err = r.KubeClient.List(ctx, specialresources)
```

controllers/specialresourcepreflight_controller.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+
+	for _, sr := range specialresources.Items {
+		if sr.Name == "special-resource-preamble" {
+			continue
+		}
+		if sr.GetDeletionTimestamp() != nil {
+			r.Log.Info("CR is marked for deletion, skipping preflight", "name", sr.Name)
+			continue
+		}
```
qbarrand, line 131: I think this check is redundant - we already made it in the `Reconcile` method.

internal/controllers/state/statusupdater.go
```diff
@@ -78,3 +80,14 @@ func (su *statusUpdater) SetAsErrored(ctx context.Context, sr *v1beta1.SpecialRe
...
+
+func (su *statusUpdater) SetVerificationStatus(ctx context.Context,
+	srp *v1beta1.SpecialResourcePreflight,
+	status *v1beta1.SROStatus,
+	verificationStatus string,
+	message string) error {
+	status.VerificationStatus = verificationStatus
+	status.StatusReason = message
+	status.LastTransitionTime = metav1.NewTime(time.Now())
+	return su.kubeClient.StatusUpdate(ctx, srp)
```
qbarrand, line 92: I think it would be safer to patch here.

controllers/specialresourcepreflight_controller.go
```diff
@@ -0,0 +1,214 @@
...
+	sroStatus := r.getPreflightSROStatus(srp, sroName)
+	errUpdate := r.StatusUpdater.SetVerificationStatus(ctx, srp, sroStatus, verificationStatus, message)
+	if errUpdate != nil {
+		r.Log.Error(err, "failed to update the status of SRO in preflight", "name", sroName)
+	}
+}
+
+func (r *SpecialResourcePreflightReconciler) presetStatusesForCRs(ctx context.Context, specialresources *srov1beta1.SpecialResourceList, srp *srov1beta1.SpecialResourcePreflight) error {
+	for _, sr := range specialresources.Items {
+		if sr.Name == "special-resource-preamble" {
```
qbarrand, line 174: Didn't we drop the preamble from the entire codebase already?

controllers/specialresourcepreflight_controller.go
```diff
@@ -0,0 +1,214 @@
...
+func (r *SpecialResourcePreflightReconciler) checkPreflightCompletion(ctx context.Context, name, namespace string) (bool, error) {
+	srp := srov1beta1.SpecialResourcePreflight{}
+	err := r.KubeClient.Get(ctx, types.NamespacedName{Namespace: namespace, Name: name}, &srp)
+	if err != nil {
+		return false, err
+	}
+
+	for _, sroStatus := range srp.Status.SROStatuses {
+		if sroStatus.VerificationStatus != srov1beta1.VerificationTrue {
+			r.Log.Info("at least one SRO CR is not veririfed yet", "name", sroStatus.Name, "status", sroStatus.VerificationStatus)
```
qbarrand, line 198:
```suggestion
			r.Log.Info("at least one SRO CR is not verified yet", "name", sroStatus.Name, "status", sroStatus.VerificationStatus)
```

api/v1beta1/specialresourcepreflight_types.go
```diff
@@ -0,0 +1,112 @@
...
+	// Namespace describes in which namespace the chart will be installed.
+	// +kubebuilder:validation:Required
+	Namespace string `json:"namespace"`
+
+	// UpdateImage describe the OCP image that all SRO need to be checked against.
+	// +kubebuilder:validation:Required
+	UpdateImage string `json:"updateImage"`
+}
+
+type SROStatus struct {
```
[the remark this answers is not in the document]
qbarrand: Applies to many fields and method / function names, IMHO.

controllers/specialresourcepreflight_controller.go
```diff
@@ -0,0 +1,214 @@
...
+	StatusUpdater state.StatusUpdater
+	KubeClient    clients.ClientsInterface
+}
+
+func (r *SpecialResourcePreflightReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&srov1beta1.SpecialResourcePreflight{}).
+		WithOptions(controller.Options{
+			MaxConcurrentReconciles: 1,
+		}).
```
qbarrand, line 57: I think it's already 1 by default?

controllers/specialresourcepreflight_controller.go
```diff
@@ -0,0 +1,214 @@
...
+	if err != nil {
+		log.Error(err, "runPreflightValidation failed")
+		return reconRes, err
+	}
+
+	if reconCompleted {
+		log.Info("Reconciliation successful for preflight", "name", srp.Name)
+		return ctrl.Result{}, nil
+	}
+	return ctrl.Result{RequeueAfter: time.Second * 60}, nil
```
qbarrand, line 96: One minute sounds arbitrary. Could we make this a setting?

api/v1beta1/specialresourcepreflight_types.go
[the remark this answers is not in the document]
yevgeny-shnaidman: Fixed

api/v1beta1/specialresourcepreflight_types.go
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

controllers/specialresourcepreflight_controller.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+func (r *SpecialResourcePreflightReconciler) runPreflightValidation(ctx context.Context, srp *srov1beta1.SpecialResourcePreflight) (bool, error) {
+	specialresources := &srov1beta1.SpecialResourceList{}
+
+	r.Log.Info("YEV - ClusterUpgradeReconciler start runPreflightValidation")
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

api/v1beta1/specialresourcepreflight_types.go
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed. Changed to PreflightValidation

controllers/specialresourcepreflight_controller.go
yevgeny-shnaidman: Lets leave all this changes for the logging epic

pkg/preflight/preflight.go
yevgeny-shnaidman: fixed

api/v1beta1/specialresourcepreflight_types.go
yevgeny-shnaidman: fixed

api/v1beta1/specialresourcepreflight_types.go
yevgeny-shnaidman: this field contains a OCP image, that will be used in the next upgrade, and the SR CRs will be verified for this image. Re-wording is welcomed :)

controllers/specialresourcepreflight_controller.go
yevgeny-shnaidman: not yet. I think we have a task somewhere, but we have not got to it yet

controllers/specialresourcepreflight_controller.go
yevgeny-shnaidman: fixed

controllers/specialresourcepreflight_controller.go
yevgeny-shnaidman: mm, not sure about it, but i also prefer to mention is explicitly, will make it easier to debug/change

controllers/specialresourcepreflight_controller.go
yevgeny-shnaidman: a setting by operator configuration or user configuration?

controllers/specialresourcepreflight_controller.go
yevgeny-shnaidman: all the logging will be checked during logging epic

controllers/specialresourcepreflight_controller.go
yevgeny-shnaidman: fixed

controllers/specialresourcepreflight_controller.go
yevgeny-shnaidman: in Reconcile method we check the PreflightValidation deletion, here we are checking the deletion of the SR CR

internal/controllers/state/statusupdater.go
yevgeny-shnaidman: Why? What is the difference between 2 of them?

controllers/preflightvalidation_controller.go
```diff
@@ -0,0 +1,210 @@
...
+
+	runInfo, err := r.PreflightAPI.PrepareRuntimeInfo(ctx, pv.Spec.UpdateImage)
+	if err != nil {
+		r.Log.Info("Failed to get runtime info from image", "image", pv.Spec.UpdateImage)
+		return false, err
+	}
+
+	err = r.KubeClient.List(ctx, specialresources)
+	if err != nil {
+		r.Log.Error(err, "Failed to get list of all SRs, error", "error")
```
pmtk, line 111:
```suggestion
		r.Log.Error(err, "Failed to get list of all SRs")
```

controllers/preflightvalidation_controller.go
```diff
@@ -0,0 +1,210 @@
...
+		return false, fmt.Errorf("failed to get list of all SRs, error %s", err)
```
pmtk, line 112:
```suggestion
		return false, fmt.Errorf("failed to get list of all SRs: %s", err)
```

api/v1beta1/preflightvalidation_types.go
```diff
@@ -0,0 +1,109 @@
...
+	// Debug enables additional logging.
+	// +kubebuilder:validation:Optional
+	Debug bool `json:"debug"`
+
+	// UpdateImage describe the OCP image that all SR CRs need to be checked against.
+	// +kubebuilder:validation:Required
+	UpdateImage string `json:"updateImage"`
+}
+
+type CRStatus struct {
```
pmtk, line 44: Why not `SRStatus`? It is a status of a Special Resource preflight validation for specific OpenShift release image. _Custom resource_ is a much wider term

controllers/preflightvalidation_controller.go
yevgeny-shnaidman: fixed

controllers/preflightvalidation_controller.go
yevgeny-shnaidman: fixed

pmtk:
looks good to me

api/v1beta1/specialresourcepreflight_types.go
qbarrand: Sure, this fits the logging epic better. But once a field is in the CRD, it's not easy to remove it 🙂

api/v1beta1/specialresourcepreflight_types.go
qbarrand: Got it 🙂 the "SRO" bit confused me. The new version is much clearer.

controllers/specialresourcepreflight_controller.go
qbarrand: What about that one @yevgeny-shnaidman?

controllers/specialresourcepreflight_controller.go
qbarrand: Right, sorry for the confusion.

internal/controllers/state/statusupdater.go
qbarrand: `Update` would replace the entire object, which `Patch` would figure out the changes and only apply those. It is helpful to avoid race conditions on object updates.

controllers/specialresourcepreflight_controller.go
qbarrand: I guess operator configuration would anyway come from the user? IMHO it would make sense to have e.g. a command-line flag for that value, with a default to one minute.

api/v1beta1/specialresourcepreflight_types.go
yevgeny-shnaidman: Since the logging epic i just around the corner, the CRD will not be yet "official" , so we will be able to change ( or , at the very least, to ignore that field)

controllers/specialresourcepreflight_controller.go
yevgeny-shnaidman: i don't think this is something that user should be aware of. I think we just need to define a constant, that can be adjusted later, once we gather enough data from preflight usage by customers.

controllers/specialresourcepreflight_controller.go
yevgeny-shnaidman: i looked at the code of the filter package closely, and it has become a mess: the same predicates should support 3 controller for 3 different type, and in addition filtering all the owned object of the CR. So, i suggest we push this PR as is, and i have opened a Jira for code refactoring (MGMT-9993) and for updating the preflight controller , once the code is refactored (MGMT-9994)

internal/controllers/state/statusupdater.go
yevgeny-shnaidman: fixed

controllers/specialresourcepreflight_controller.go
qbarrand: It can be as simple as https://github.com/qbarrand/oot-operator/blob/main/controllers/predicates/predicates.go#L21

yevgeny-shnaidman: yes, but it would still go into the filter package. So i would prefer that it will be refactored properly, and then the controllers will be updated. We also have a code in the specialresource reconciliation that can be handled by a proper filtering too, so this work can be done in a separate PR

controllers/specialresourcepreflight_controller.go
yevgeny-shnaidman: currently moved it to a constant

PR #169, opened 2022-04-17
MGMT-10014: preflight should handle multiple daemonsets and buildconf…

yevgeny-shnaidman:
…ig in a chart

This PR adds the following abilities to the preflight:
1) find all the daemon sets and buildconfig in the chart
2) ignore daemonset that are not drvier containers
   annotation specialresource.openshift.io/state is not set to driver-container
3) ignore daemonsets that have a paired build configs meaning:
   annotation specialresource.openshift.io/driver-container-vendor with the same value
   is present in buildconfig and in daemonset
4) run validation on the rest of the daemonsets
5) in case even one of the daemonset fails validation - the whole CR validation is failed


yevgeny-shnaidman:
@assign @pmtk

pkg/preflight/preflight.go
```diff
@@ -131,38 +131,75 @@ func (p *preflight) processFullChartTemplates(ctx context.Context,
...
 func (p *preflight) handleYAMLsCheck(ctx context.Context, yamlsList string, upgradeKernelVersion string) (bool, string, error) {
 	objList, err := p.resourceAPI.GetObjectsFromYAML([]byte(yamlsList))
 	if err != nil {
 		p.log.Error(err, "failed to extract object from chart yaml list during preflight")
 		return false, "Failed to extract object from chart yaml list during preflight", err
 	}
 
+	daemonSetList, buildConfigPresent, err := p.getRelevantDaemonSets(objList)
+	if err != nil {
+		return false, "Error while trying to parse BuildConfigs and DaemonSets", err
```
ybettan, line 143: Why do we need that bool? looks like it is always `false` when an error occurred and `true` otherwise.
Should we remove it?

yevgeny-shnaidman: look at line 154, it returns the result of the validation per specific DaemonSet

ybettan: Sorry, I missed that.

pkg/preflight/preflight.go
```diff
@@ -131,38 +131,75 @@ func (p *preflight) processFullChartTemplates(ctx context.Context,
...
-			p.log.Info("preflight: buildconfig related to daemonset, skipping image verification")
-			return true, VerificationStatusReasonBuildConfigPresent, nil
+			buildConfigPresent = true
+			annotations := obj.GetAnnotations()
+			driverContainerToken, found := annotations["specialresource.openshift.io/driver-container-vendor"]
+			if found {
+				buildConfigMap[driverContainerToken] = struct{}{}
+			}
 		}
 		if kind == "DaemonSet" {
```
ybettan, line 176: Maybe we can use `switch-case` or `else if` instead of another `if`?

pkg/preflight/preflight.go
```diff
@@ -131,38 +131,75 @@ func (p *preflight) processFullChartTemplates(ctx context.Context,
...
 		kind := obj.GetKind()
 		if kind == "BuildConfig" {
-			// no more need to check daemons set, build config is present
-			p.log.Info("preflight: buildconfig related to daemonset, skipping image verification")
-			return true, VerificationStatusReasonBuildConfigPresent, nil
+			buildConfigPresent = true
+			annotations := obj.GetAnnotations()
+			driverContainerToken, found := annotations["specialresource.openshift.io/driver-container-vendor"]
+			if found {
+				buildConfigMap[driverContainerToken] = struct{}{}
```
ybettan, line 173: Don't we want to set it to `obj`?

pkg/preflight/preflight.go
```diff
@@ -131,38 +131,75 @@ func (p *preflight) processFullChartTemplates(ctx context.Context,
...
+			}
+			state, found := daemonSet.Annotations["specialresource.openshift.io/state"]
+			if found && state == "driver-container" {
+				tempDSList = append(tempDSList, &daemonSet)
+			}
 		}
 	}
-	return true, VerificationStatusReasonNoDaemonSet, nil
+	resultList := []*k8sv1.DaemonSet{}
+	for _, item := range tempDSList {
```
ybettan, line 190: Why do we need that second loop, can't we have that logic in the first loop if this is a DS object?

pkg/preflight/preflight.go
yevgeny-shnaidman: no, we have no other use for BuildConfig except for its annotation, so why keep the obj

pkg/preflight/preflight.go
yevgeny-shnaidman: in the first loop we are just creating the daemonset slice and buildconfig map. Since we don't know the exact sequence, we cannot rely on daemonset being after buildconfigs. The second loop removes from the daemonset slice all the daemons that have a paired buildconfig

pkg/preflight/preflight.go
yevgeny-shnaidman: fixed

pkg/preflight/preflight.go
ybettan: The default value of `buildConfigMap[driverContainerToken]` is an empty struct, see https://go.dev/play/.
Why do we need to set it explicitly?

yevgeny-shnaidman: when you add a new key, you should always add a value for that key. I don't know any other way in golang to add a key other then:
map[key]=value

PR #173, opened 2022-04-27
Bug 2078781: fix preflight validation for multi-arch OCP images

yevgeny-shnaidman:
since 4.11, the OCP images are multi-arch images, which causes
preflight validation to crush due to incorrect image structure parsing.
This PR introduces the following changes in the registry package GetLayersDigests
function, which parses the OCP images:
1) get the manifest stream for OCP images
2) check the media type of the image
3) in case media type is manifest list (multi-arch):
   a) parse the manifest list to obtain the digest of the image for the running platform
   b) repace the previously obtained manifest with the manifest of the images
      suitable for the platform (found in step a)


pkg/registry/registry.go
```diff
@@ -187,8 +188,8 @@ func (r *registry) ReleaseImageMachineOSConfig(layer v1.Layer) (string, error) {
 	return "", fmt.Errorf("failed to find machine-os-content in the release-manifests/image-references")
 }
 
-func (r *registry) GetLayersDigests(ctx context.Context, entry string) (string, []string, []crane.Option, error) {
-	registry, err := r.registryFromImageURL(entry)
+func (r *registry) GetLayersDigests(ctx context.Context, image string) (string, []string, []crane.Option, error) {
```
ybettan, line 191: Now that this method has changed, should we add an additional test for it that passes a multi-arch image?

pkg/registry/registry.go
```diff
@@ -200,39 +201,30 @@ func (r *registry) GetLayersDigests(ctx context.Context, entry string) (string,
...
+		return "", nil, nil, fmt.Errorf("image url %s is not valid, does not contain hash or tag", image)
 	}
 
 	var registryAuths []crane.Option
 	if auth.Auth != "" {
 		registryAuths = append(registryAuths, crane.WithAuth(authn.FromConfig(authn.AuthConfig{Username: auth.Email, Auth: auth.Auth})))
 	}
 
-	manifest, err := crane.Manifest(entry, registryAuths...)
+	manifest, err := r.getManifestStreamFromImage(image, repo, registryAuths)
```
ybettan, line 219: Isn't `repo` part of `image` string?

pkg/registry/registry.go
```diff
@@ -241,6 +233,50 @@ func (r *registry) GetLayerByDigest(repo string, digest string, auth []crane.Opt
 	return crane.PullLayer(repo+"@"+digest, auth...)
 }
 
+func (r *registry) getManifestStreamFromImage(image, repo string, options []crane.Option) ([]byte, error) {
+	manifest, err := crane.Manifest(image, options...)
+	if err != nil {
+		return nil, err
```
ybettan, line 239: You leave the "wrapping error" task to a separate PR?

pkg/registry/registry.go
yevgeny-shnaidman: no, image contains repo + sha2 + hash

pkg/registry/registry.go
yevgeny-shnaidman: yes, there is already on in the logging epic

pkg/registry/registry.go
```diff
@@ -241,6 +233,50 @@ func (r *registry) GetLayerByDigest(repo string, digest string, auth []crane.Opt
...
+	}
+
+	release := unstructured.Unstructured{}
+	if err = json.Unmarshal(manifest, &release.Object); err != nil {
+		return nil, err
+	}
+
+	imageMediaType, mediaTypeFound, err := unstructured.NestedString(release.Object, "mediaType")
+	if err != nil || !mediaTypeFound {
+		return nil, err
```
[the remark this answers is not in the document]
yevgeny-shnaidman: Great catch, thanks! Fixed

PR #177, opened 2022-05-01
Enhancing returned errors for pkg/lifecycle.

ybettan:
This is part of the logging enhancement effort which says that:
  * all errors should be wrapped with its context
  * error logging should only be done at the reconciliation loop level
  * info and warning logs can be everywhere in the code

### DIFF
--- a/pkg/lifecycle/lifecycle.go
+++ b/pkg/lifecycle/lifecycle.go
@@ -2,6 +2,7 @@ package lifecycle
 
 import (
 	"context"
+	"fmt"
 	"os"
 
 	"github.com/openshift/special-resource-operator/pkg/clients"
@@ -40,7 +41,9 @@ func (l *lifecycle) GetPodFromDaemonSet(ctx context.Context, key types.Namespace
 
 	err := l.kubeClient.Get(ctx, key, ds)
 	if apierrors.IsNotFound(err) || err != nil {
-		ctrl.LoggerFrom(ctx).Error(err, "Failed to get DaemonSet", "key", key)
+		if err != nil {
+			ctrl.LoggerFrom(ctx).Info(utils.WarnString("Failed to get DaemonSet"), "key", key, "error", err)
+		}
 		return &v1.PodList{}
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/lifecycle/lifecycle.go
```diff
@@ -40,7 +41,7 @@ func (l *lifecycle) GetPodFromDaemonSet(ctx context.Context, key types.Namespace
 
 	err := l.kubeClient.Get(ctx, key, ds)
 	if apierrors.IsNotFound(err) || err != nil {
-		ctrl.LoggerFrom(ctx).Error(err, "Failed to get DaemonSet", "key", key)
+		ctrl.LoggerFrom(ctx).Info(utils.WarnString("Failed to get DaemonSet"), "key", key, "error", err)
```
ybettan, line 44: @yevgeny-shnaidman @qbarrand This is what I meant in the [design doc](https://docs.google.com/document/d/1CHctpDQrjAABhiJ6GpEVhc8dfwBiXYJY56zl4ytoqWs/edit?disco=AAAAXyLIa4E).
WDYT?

ybettan: BTW by adding the (maybe wrapped) error itself to Warn logs we may log the full error stack multiple times. Is this something we want?

yevgeny-shnaidman: since the in the warning we don't return the error, then iit should not be an issue

pkg/lifecycle/lifecycle.go
```diff
@@ -52,7 +53,7 @@ func (l *lifecycle) GetPodFromDeployment(ctx context.Context, key types.Namespac
 
 	err := l.kubeClient.Get(ctx, key, dp)
 	if apierrors.IsNotFound(err) || err != nil {
-		ctrl.LoggerFrom(ctx).Error(err, "Failed to get Deployment", "key", key)
+		ctrl.LoggerFrom(ctx).Info(utils.WarnString("Failed to get Deployment"), "key", key, "error", err)
```
yevgeny-shnaidman, line 56: i wonder if we should split all those "ifs" into 2 parts:
1) error of not found
2) all other errors
and log the warning only on all other errors.  It is perfectly normal for deployments/ds/ss etc' not to exists, since it might take time to create them, and we don't want to clutter our log with warnings. WDYT?

ybettan: I don't mind. Will fix.

ybettan: Fixed.

PR #188, opened 2022-05-03
Enhancing returned errors for specialresourcemodule controller.

ybettan:
This is part of the logging enhancement effort which says that:
  * all errors should be wrapped with its context
  * error logging should only be done at the reconciliation loop level
  * info and warning logs can be everywhere in the code

### DIFF
--- a/controllers/specialresourcemodule_controller.go
+++ b/controllers/specialresourcemodule_controller.go
@@ -46,6 +46,7 @@ import (
 	"github.com/openshift/special-resource-operator/pkg/registry"
 	"github.com/openshift/special-resource-operator/pkg/resource"
 	sroruntime "github.com/openshift/special-resource-operator/pkg/runtime"
+	"github.com/openshift/special-resource-operator/pkg/utils"
 	"github.com/openshift/special-resource-operator/pkg/watcher"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -64,8 +65,9 @@ const (
 	minDelaySRM = 100 * time.Millisecond
 	maxDelaySRM = 3 * time.Second
 
-	SRMgvk        = "SpecialResourceModule"
-	SRMOwnedLabel = "specialresourcemodule.openshift.io/owned"
+	SRMgvk          = "SpecialResourceModule"
+	SRMOwnedLabel   = "specialresourcemodule.openshift.io/owned"
+	upgradesInfoURL = "https://api.openshift.com/api/upgrades_info/v1/graph"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


controllers/specialresourcemodule_controller.go
```diff
@@ -145,47 +146,47 @@ func getImageFromVersion(entry string) (string, error) {
 	res := versionRegex.FindStringSubmatch(entry)
 	full, major, minor := res[0], res[1], res[2]
 	var imageURL string
+	const upgradesInfoURL = "https://api.openshift.com/api/upgrades_info/v1/graph"
```
yevgeny-shnaidman, line 149: const should be outside of the function definition

controllers/specialresourcemodule_controller.go
```diff
@@ -145,47 +146,47 @@ func getImageFromVersion(entry string) (string, error) {
...
-			http.DefaultTransport,
-		)
+		config := transport.Config{
+			UserAgent: rest.DefaultKubernetesUserAgent() + "(release-info)",
+		}
+		roundTripper := http.DefaultTransport
+		rt, err := transport.HTTPWrappersForConfig(&config, roundTripper)
 		if err != nil {
-			return "", err
+			return "", fmt.Errorf("failed to wrap config %v with roundTripper %v: %w", config, roundTripper, err)
```
yevgeny-shnaidman, line 157: do we want to start printing the whole struct in the error? won't it make  trace a little harder to read? I think that in case we want to print the whole struct , then we should do it via debug

controllers/specialresourcemodule_controller.go
```diff
@@ -145,47 +146,47 @@ func getImageFromVersion(entry string) (string, error) {
...
 				req, err := http.NewRequest("GET", u.String(), nil)
 				if err != nil {
-					return err
+					return fmt.Errorf("failed to create HTTP request 'GET %s': %w", u.String(), err)
 				}
 				req.Header.Set("Accept", "application/json")
 				resp, err := cl.Do(req)
 				if err != nil {
-					return err
+					return fmt.Errorf("failed to create HTTP request %v: %w", req, err)
```
yevgeny-shnaidman, line 174: same comment as above

controllers/specialresourcemodule_controller.go
```diff
@@ -145,47 +146,47 @@ func getImageFromVersion(entry string) (string, error) {
...
 				case http.StatusOK:
 				default:
 					_, _ = io.Copy(ioutil.Discard, resp.Body)
-					return fmt.Errorf("unable to retrieve image. status code %d", resp.StatusCode)
+					return fmt.Errorf("unable to retrieve image, status code %d", resp.StatusCode)
 				}
 				data, err := ioutil.ReadAll(resp.Body)
 				if err != nil {
-					return err
+					return fmt.Errorf("unable to read reponse %v: %w", resp, err)
```
yevgeny-shnaidman, line 185: same comment as above

controllers/specialresourcemodule_controller.go
```diff
@@ -145,47 +146,47 @@ func getImageFromVersion(entry string) (string, error) {
...
 				}
 				var versions versionGraph
 				if err := json.Unmarshal(data, &versions); err != nil {
-					return err
+					return fmt.Errorf("unalbe to unmarshal json %s: %w", data, err)
```
yevgeny-shnaidman, line 189: same comment as above regarding printing the whole json string

controllers/specialresourcemodule_controller.go
```diff
@@ -262,7 +264,7 @@ func (r *SpecialResourceModuleReconciler) filterResources(selectors []srov1beta1
 		for _, obj := range objs {
 			candidates, err := watcher.GetJSONPath(selector.Path, obj)
 			if err != nil {
-				return filteredObjects, err
+				return []unstructured.Unstructured{}, fmt.Errorf("failed to get json path %s for obj %v: %w", selector.Path, obj, err)
```
yevgeny-shnaidman, line 267: same comment as above regarding object

controllers/specialresourcemodule_controller.go
```diff
@@ -318,36 +320,36 @@ func (r *SpecialResourceModuleReconciler) getOCPVersions(ctx context.Context, wa
 			if k8serrors.IsNotFound(err) {
 				continue
 			}
-			return nil, err
+			return nil, fmt.Errorf("failed to get %s/%s in namespace %s: %w", resource.Kind, resource.Name, resource.Namespace, err)
 		}
 		objs, err = r.filterResources(resource.Selector, objs)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("failed to filter resource %v with filter %v: %w", objs, resource.Selector, err)
```
yevgeny-shnaidman, line 327: same comment as above regarding object

controllers/specialresourcemodule_controller.go
```diff
@@ -368,7 +370,7 @@ func (r *SpecialResourceModuleReconciler) reconcileChart(ctx context.Context, sr
 	result := make([]string, 0)
 	c, err := r.Helmer.Load(srm.Spec.Chart)
 	if err != nil {
-		return result, err
+		return []string{}, fmt.Errorf("could not load helm chart %v: %w", srm.Spec.Chart, err)
```
yevgeny-shnaidman, line 373: same comment regrading the rest of the changes

controllers/specialresourcemodule_controller.go
ybettan: Why is that? if it is only used in that function, why should we increase its scope?

controllers/specialresourcemodule_controller.go
ybettan: Fixed.

controllers/specialresourcemodule_controller.go
ybettan: Fixed.

controllers/specialresourcemodule_controller.go
ybettan: Fixed.

controllers/specialresourcemodule_controller.go
ybettan: Fixed.

controllers/specialresourcemodule_controller.go
ybettan: Fixed.

controllers/specialresourcemodule_controller.go
ybettan: Fixed.

controllers/specialresourcemodule_controller.go
ybettan: Fixed.

controllers/specialresourcemodule_controller.go
yevgeny-shnaidman: in GO you usually do  not define const in the function scope

ybettan: Fixed.

ybettan:
@yevgeny-shnaidman PTAL

ybettan:
Created a follow up [task](https://issues.redhat.com/browse/MGMT-10259) for adding info logs to this controller.

PR #189, opened 2022-05-03
MGMT-10114: Enhance logging for package cluster

yevgeny-shnaidman:


pkg/cluster/cluster.go
```diff
@@ -26,8 +26,6 @@ type Cluster interface {
 	OSImageURL(context.Context) (string, error)
 	OperatingSystem(*corev1.NodeList) (string, string, string, error)
 	GetDTKImages(context.Context) ([]string, error)
-	NextUpgradeVersion(context.Context) (string, error)
-	IsClusterInUpgrade(context.Context) (bool, error)
```
ybettan, line 30: 1. How this is part of the logging enhancement?
2. Don't we need it anymore? why are we removing it?
3. Should be change the commit message and PR title and description as this is the main logical change in the PR

yevgeny-shnaidman: This is not a part of logging, but i thought to add it here, and not to open a different PR for that , since it is a small change. We don't use them anymore, they were part of previous version of preflight solution

yevgeny-shnaidman: i will add it as a step in a commit

PR #194, opened 2022-05-06
Bug 2070172: Use the SpecialResource's name as Helm release

qbarrand:
This PR makes `pkg/helmer` code use the SpecialResource's name as the Helm release name. This avoids collisions when two SpecialResources are targeting the same chart.


ybettan:
@qbarrand You have some linting issues.

ybettan:
@qbarrand Please squash the PR and update the commit message to be descriptive as the PR description.

PR #204, opened 2022-05-12
MGMT-10115: Enhance logging for package filter

yevgeny-shnaidman:


ybettan:
When do we decide what level of verbosity we want in the logs?
1. when we run `oc log --verbosity` (or similar)
2. when deploying SRO
?

PR #206, opened 2022-05-12
MGMT-10128: Enhance logging for package watcher

yevgeny-shnaidman:


pkg/watcher/watcher.go
```diff
@@ -323,12 +317,12 @@ func GetJSONPath(path string, obj unstructured.Unstructured) ([]string, error) {
...
 			return nil, nil
 		}
-		return nil, err
+		return nil, fmt.Errorf("failed to lookup object %s/%s: %w", obj.GetNamespace(), obj.GetName(), err)
 	}
 	switch reflect.TypeOf(match).Kind() {
 	case reflect.Slice:
 		if res, ok := match.([]interface{}); !ok {
-			return nil, errors.New("error converting result to string")
+			return nil, fmt.Errorf("error converting result to string")
```
ybettan, line 325:
```suggestion
			return nil, errors.New("error converting result to string")
```

ybettan:
/hold
Just to make sure it is not getting merged (although CI is broken) while reviewing.

pkg/watcher/watcher.go
yevgeny-shnaidman: why? there is not difference between errors.New and fmt.Errorf. Since we are using fmt.Errorf in other cases, let's continue with it

yevgeny-shnaidman: both of them create error.  so, why use 2 methods, and 2 imports, instead of one? i have looked for some kind of programmatic reason to prefer errors.New, and found none

yevgeny-shnaidman: fixed

pkg/watcher/watcher.go
```diff
@@ -339,7 +333,7 @@ func GetJSONPath(path string, obj unstructured.Unstructured) ([]string, error) {
 	case reflect.String:
 		return []string{match.(string)}, nil
 	}
-	return nil, errors.New("unsupported result")
+	return nil, fmt.Errorf("unsupported result")
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

PR #207, opened 2022-05-12
MGMT-10259: Add info logs for specialresourcemodule controller

pacevedom:


controllers/specialresourcemodule_controller.go
```diff
@@ -250,7 +257,7 @@ func (r *SpecialResourceModuleReconciler) getAllResources(kind, apiVersion, name
 	obj.SetName(name)
 	key := client.ObjectKeyFromObject(&obj)
 	if err := r.KubeClient.Get(context.Background(), key, &obj); err != nil {
-		return []unstructured.Unstructured{}, fmt.Errorf("failed to get %s/%s in namespace %s: %w", kind, name, namespace, err)
+		return []unstructured.Unstructured{}, fmt.Errorf("failed to get '%s/%s' in namespace '%s': %w", kind, name, namespace, err)
```
yevgeny-shnaidman, line 260: lets make it: "fmt.Errorf("failed to get object kind %s '%s/%s' ': %w", kind, name, namespace, err)
our convention, in case we print kuberentes object is <namespace/name>

yevgeny-shnaidman:
Also, i think we should add more logs to the Reconcile function that will show the steps of each reconcilaition

controllers/specialresourcemodule_controller.go
pacevedom: Done

PR #208, opened 2022-05-15
Bug 2084732: fix SRO CR finalization in case of an upgrade

yevgeny-shnaidman:
In case of cluster(SRO) upgrade from 4.9 to 4.10 or 4.11,
an SRO CR will contain to finalizers strings, since the finalization
string has changed between 4.9 and 4.10. Current code only removes
current finalization atring, which means that in case of an upgrade
finalization string of 4.9 is left on the CR and it is never deleted
from etcd. This PR will always try to remove also the 4.9 finalization
string together with current finalization string

### DIFF
--- a/internal/controllers/finalizers/specialresource_test.go
+++ b/internal/controllers/finalizers/specialresource_test.go
@@ -1,4 +1,4 @@
-package finalizers_test
+package finalizers
 
 import (
 	"context"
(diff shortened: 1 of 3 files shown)


internal/controllers/finalizers/specialresource_test.go
```diff
@@ -9,7 +9,7 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"github.com/openshift/special-resource-operator/api/v1beta1"
-	"github.com/openshift/special-resource-operator/internal/controllers/finalizers"
+	//"github.com/openshift/special-resource-operator/internal/controllers/finalizers"
```
ybettan, line 12: I guess we don't need that comment anymore?!

yevgeny-shnaidman: fixed

PR #211, opened 2022-05-19
Bug 2083514: Add resource removal capabilities

pacevedom:


controllers/specialresource.go
```diff
@@ -327,7 +327,7 @@ func (r *SpecialResourceReconciler) createSpecialResourceFrom(ctx context.Contex
 }
 
 func (r *SpecialResourceReconciler) removeSpecialResource(ctx context.Context, sr *srov1beta1.SpecialResource) error {
-	if err := r.Finalizer.Finalize(ctx, sr); err != nil {
+	if err := r.Finalizer.RemoveResources(ctx, SROwnedLabel, sr); err != nil {
```
ybettan, line 330:
```suggestion
	if err := r.Finalizer.RemoveResources(ctx, SROOwnedLabel, sr); err != nil {
```

internal/controllers/finalizers/specialresource.go
```diff
@@ -150,6 +153,48 @@ func (srf *specialResourceFinalizer) finalizeSpecialResource(ctx context.Context
 			}
 		}
 	}
+	return nil
+}
+
+func (srf *specialResourceFinalizer) RemoveResources(ctx context.Context, ownedLabel string, sr *v1beta1.SpecialResource) error {
```
ybettan, line 159: I am wondering why we need to make this deletion manually. By adding an `owner` to and object, you ensure that once that owner object is deleted then all its owned objects are deleted with it. It is a k8s feature.

ybettan:
@pacevedom Please add a commit message explaining the change and why it is needed. Same for PR description.

PR #212, opened 2022-05-19
Draft: use host config & CA certificates when pulling images

qbarrand:
This PR wraps `crane` calls into a type that reads the system's `registries.conf` to resolve the image name to the desired registry.
In addition, the wrapper sets up useful `crane` options:
- optional auth, if it is configured in the pull secret;
- new transport with the pod's and the host's CA certificates, and;
- the current reconciliation's context.


ybettan:
/close
Closing this PR in favor of https://github.com/openshift/special-resource-operator/pull/226

PR #213, opened 2022-05-23
Bug 2083486: Support for operator Force/Managed/Removed/Unmanaged states

pacevedom:
Support for operator Force/Managed/Removed/Unmanaged states


bthurber (approved):
/lgtm

PR #215, opened 2022-05-24
Bug 2089719: Add arch to kernel versions

pacevedom:
Solves [BZ 2089719](https://bugzilla.redhat.com/show_bug.cgi?id=2089719) by adding the arch when its missing from the release info in DTK (happens in older versions). This renders builds not working when using the `kernelFullVersion` or `rKernelFullVersion` variables in recipes.

### DIFF
--- a/controllers/specialresourcemodule_controller.go
+++ b/controllers/specialresourcemodule_controller.go
@@ -25,6 +25,7 @@ import (
 	"net/http"
 	"net/url"
 	"regexp"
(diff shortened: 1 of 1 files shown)


ybettan:
@pacevedom You need to repush this PR in order to consume https://github.com/openshift/release/pull/28868.
If you do so, then please also update the commit message and PR description (not only PR description) to contains an explanation on why such change is needed.

PR #217, opened 2022-05-25
Bug 2089344: fix the Reconcilition function logs

yevgeny-shnaidman:
This PR moves the resulting logging of errors and warning from the
top Reconcile function to the SpecialResourcesReconcile function,
and simplifies the Reconcile function


pacevedom (approved):
/lgtm

PR #219, opened 2022-06-02
Bug 2092572: Add namespace creation to acm-simple-kmod

pacevedom:
Adds the namespace creation to the policy for the spokes.


ybettan:
@pacevedom Commit message is missing. Can you please add it to the commit?

pacevedom:
PTAL @pmtk

PR #222, opened 2022-06-06
Bug 2088324: Some manifests still have KVC references

enriquebelarte:
Manifests from bundle and example in charts still refer to kvc-simple-kmod which leads to errors as[ kvc-simple-kmod repo](https://github.com/openshift-psap/kvc-simple-kmod) is no longer used by SRO in favor of [simple-kmod](https://github.com/openshift-psap/simple-kmod) which does not have [KVC](https://github.com/kmods-via-containers/kmods-via-containers) dependencies.


ybettan:
@enriquebelarte Can you please copy the PR description inside the commit message?

PR #223, opened 2022-06-06
Bug 2093357: hub and spokes: add OCP version retrieval and fix ICE driver

pacevedom:
Add functionality to extract OCP version from images and use it as
a helm variable.
Tweak ICE driver example to fix bug 2093357

### DIFF
--- a/charts/example/acm-ice-0.0.1/templates/0001-buildconfig.yaml
+++ b/charts/example/acm-ice-0.0.1/templates/0001-buildconfig.yaml
@@ -33,8 +33,8 @@ apiVersion: build.openshift.io/v1
 kind: BuildConfig
 metadata:
   labels:
-    app: {{ printf "%s-%s" .Values.specialResourceModule.metadata.name .Values.kernelFullVersion | replace "." "-" | replace "_" "-" | trunc 63 }}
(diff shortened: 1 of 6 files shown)


cdvultur:
This fixes the driver problem. Still wondering if , with the new kernel ( in case of the reboot while OCP upgrades ) the old driver can be loaded or not and what happens in that case. 

If it does not load ( error our the service execution ) , will it do the same thing ?

pacevedom:
> This fixes the driver problem. Still wondering if , with the new kernel ( in case of the reboot while OCP upgrades ) the old driver can be loaded or not and what happens in that case.
> 
> If it does not load ( error our the service execution ) , will it do the same thing ?

This is a special situation that should be handled with care. If you know you are going to run into incompatibilities (I need X version of the kernel because some symbols are only there, for example) you would need to plan in advance. This means you would need to build the driver container before the actual upgrade.

What I wonder is when the `openshiftVersion` from the spoke to the hub is upgraded. If this happens before the machine config template is applied (which I doubt), then you dont have any issues.

cdvultur:
From what I see, the openshiftVersion is updated at the end of the upgrade. 

If you start upgrade, and run `oc get clusterversion` you will see the old version of the cluster in the VERSION column and the fact that the cluster is upgraded in the STATUS column. Once it reaches 100% , the VERSION is bumped UP .

PR #224, opened 2022-06-07
Bug 2096162: Use official registries for ACM examples

pacevedom:
Remove team members users referenced in registry config
from ACM SRM examples.


bthurber (approved):

pacevedom:
> @pacevedom do we have a BZ for this?

Not really, its something I realised when reviewing upcoming docs. I will create one.

PR #226, opened 2022-06-13
Adding support for disconnected clusters.

ybettan:
Disconnected clusters in OCP uses an ImageContentSourcePolicy resource
to map OCP image names to their mirror on a custom registry.

This ImageContentSourcePolicy writes mirroring configuration for the
container runtime on the nodes directly; the redirection is completely
transparent for other Kubernetes components, including SRO.

This means that the image referenced in the openshift/driver-toolkit
ImageStream cannot be directly pulled by SRO.

This PR is adding support for disconnected envs by wrapping crane
calls into a type that reads the system's registries.conf to resolve the
image name to the desired registry.

In addition, the wrapper sets up useful crane options:
  * optional auth, if it is configured in the pull secret
  * new transport with the pod's and the host's CA certificates
  * the current reconciliation's context

Co-authored-by: Yoni Bettan <yonibettan@gmail.com>


bthurber (approved):

pkg/registry/wrapper.go
```diff
@@ -0,0 +1,202 @@
...
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/openshift/special-resource-operator/pkg/clients"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var ErrNoAuthForRegistry = errors.New("no auth defined for this registry")
+
+//go:generate mockgen -source=wrapper.go -package=registry -destination=mock_wrapper_api.go
+
+type MirrorResolver interface {
```
yevgeny-shnaidman, line 28: why do we need this interface? it is only used by the crane wrapper and it is in its package. So lets either make it methods of the wrapper or, if we think that we will use it for other purposes, move it to a different package

pkg/registry/wrapper.go
```diff
@@ -0,0 +1,202 @@
...
+	// r is nil if none of the registries in registries.conf matched imageName.
+	// In that case, return that reference as pull source.
+	if r == nil {
+		return []sysregistriesv2.PullSource{{Reference: n}}, nil
+	}
+
+	return r.PullSourcesFromReference(n)
+}
+
+type CertPoolGetter interface {
```
yevgeny-shnaidman, line 64: same comment for the  CertPoolGetter: either make it a method, or move to the same package as MirrorResolver

pkg/registry/wrapper.go
ybettan: Removed the interface and replaced it with a private method for the `crateWrapper` object instead.

pkg/registry/wrapper.go
ybettan: Removed the interface and replaced it with a private method for the `crateWrapper` object instead.

ybettan:
@cdvultur Do you have a way to test that PR on disconnected envs before we merge it?

ybettan:
/hold
Remove all FIXMEs

config/manager/manager.yaml
```diff
@@ -45,7 +45,9 @@ spec:
           image: controller:latest
           name: manager
           securityContext:
-            allowPrivilegeEscalation: false
+            #FIXME:ybettan user less pervissive access
+            privileged: true
```
ybettan, line 49: @yevgeny-shnaidman - Couldn't find a more restrictive linux capability to access `/mnt/host/`tls-ca-bundle.pem

ybettan: Duplication. Resolving.

pkg/registry/wrapper.go
```diff
@@ -0,0 +1,183 @@
...
+		return "", nil, fmt.Errorf("empty pull sources for image %q: %v", imageName, err)
+	}
+
+	certPool, err := cw.systemAndHostCertPool()
+	if err != nil {
+		return "", nil, fmt.Errorf("could not create a client certificate pool: %v", err)
+	}
+
+	t := http.DefaultTransport.(*http.Transport).Clone()
+	t.TLSClientConfig.ClientCAs = certPool
```
qbarrand, line 144:
```suggestion
	t.TLSClientConfig.RootCAs = certPool
```

ybettan: Fixed.

pkg/registry/wrapper.go
```diff
@@ -0,0 +1,183 @@
...
+			return "", nil, fmt.Errorf("could not find auth for registry %s: %v", registry, err)
+		}
+	} else {
+		opts = append(opts, crane.WithAuth(auth))
+	}
+
+	// In case the image is mirrored we will get a different registry
+	newImageNameList := append([]string{registry}, strings.Split(imageName, "/")[1:]...)
+	newImageName := strings.Join(newImageNameList, "/")
+	return newImageName, opts, nil
```
qbarrand, line 164:
```suggestion
	return return ps[0].Reference.String(), opts, nil
```

PR #235, opened 2022-06-21
Bug 2099507: failure to update special-resource-lifecycle table durin…

yevgeny-shnaidman:
…g update Event

During the update event, the filtering code is trying to update the special-resource-lifecycle table
and fails with "the object has been modified" error. Since there is no retry in the filtering event
and not error returning, the flow continues, without the table being updated.
This later causes the problem with the upgrade of the daemonset with UpgradeStrategy set to OnDelete.

This PR stops relying and using a special table just for that process. special-resource-lifecycle table
is removed as also all the code that is using it. Now we use the generation label and
pod generation template label which are set and managed by the DS controller,
which provides a more reliable mechanism


pkg/poll/poll.go
```diff
@@ -365,15 +362,20 @@ func ForLifecycleAvailability(obj *unstructured.Unstructured) error {
...
-			hs, err := hash.FNV64a(pod.GetNamespace() + pod.GetName())
-			if err != nil {
-				return false, err
-			}
-			value, err := storage.CheckConfigMapEntry(hs, ins)
-			if err != nil {
-				return false, err
-			}
-			if value != "" {
+			//hs, err := hash.FNV64a(pod.GetNamespace() + pod.GetName())
```
ybettan, line 365: Do we need those comments?

pkg/filter/filter.go
```diff
@@ -284,19 +269,6 @@ func Predicate() predicate.Predicate {
...
-				ins := types.NamespacedName{
-					Namespace: os.Getenv("OPERATOR_NAMESPACE"),
-					Name:      "special-resource-lifecycle",
-				}
-				key, err := hash.FNV64a(obj.GetNamespace() + obj.GetName())
-				if err != nil {
-					warn.OnError(err)
-					return false
-				}
-				err = storage.DeleteConfigMapEntry(key, ins)
```
qbarrand, line 297: I think we can remove that function from `pkg/storage`. Nothing is calling it anymore.

pkg/poll/poll.go
```diff
@@ -347,16 +346,14 @@ func ForLifecycleAvailability(obj *unstructured.Unstructured) error {
 		return nil
 	}
 
+	annotations := obj.GetAnnotations()
+        tempGenerator := annotations[apps.DeprecatedTemplateGeneration]
```
qbarrand: Bad indentation (spaces instead of 1 tab).

pkg/poll/poll.go
yevgeny-shnaidman: fixed

pkg/resource/resource.go
```diff
@@ -433,6 +450,27 @@ func SetMetaData(obj *unstructured.Unstructured, nm string, ns string) {
...
+	if !ok {
+		return
+	}
+	annotations := obj.GetAnnotations()
+	if annotations == nil {
+		annotations = make(map[string]string)
+	}
+
+	annotations[apps.DeprecatedTemplateGeneration] = tempGeneration
+	obj.SetAnnotations(annotations)
```
qbarrand, line 471: Will this ever set the annotation? It returns without setting anything if the annotation is not already present.

pkg/lifecycle/lifecycle.go
```diff
@@ -2,98 +2,51 @@ package lifecycle
...
-	log logr.Logger
 )
 
-func init() {
-	log = zap.New(zap.UseDevMode(true)).WithName(color.Print("lifecycle", color.Green))
-}
-
 func GetPodFromDaemonSet(key types.NamespacedName) *v1.PodList {
-        ds := &appsv1.DaemonSet{}
+	ds := &appsv1.DaemonSet{}
```
qbarrand, line 16: Indentation looks a bit off in the whole PR. Should be 1 tab, no spaces.

pkg/poll/poll.go
```diff
@@ -365,15 +362,20 @@ func ForLifecycleAvailability(obj *unstructured.Unstructured) error {
...
+			//if err != nil {
+			//	return false, err
+			//}
+			//value, err := storage.CheckConfigMapEntry(hs, ins)
+			//if err != nil {
+			//	return false, err
+			//}
+			//if value != "" {
+			//	return false, nil
+			//}
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

pkg/poll/poll.go
yevgeny-shnaidman: run the fmt on all the files

pkg/lifecycle/lifecycle.go
yevgeny-shnaidman: run the fmt

pkg/resource/resource.go
yevgeny-shnaidman: the goal here is to update the obj with the annotations present in the object i the cluster. So, in case no annotations are there yet, we just return ( no need to update the obj). Once the DS controller will update the annotations, the obj will get update, and the poll loop will be able to complete

pkg/filter/filter.go
yevgeny-shnaidman: fixed

ybettan:
> Needs cherry-pick-approved label

@bthurber @cdvultur who can add this label?

bthurber:
> > Needs cherry-pick-approved label
> 
> @bthurber @cdvultur who can add this label?

QE is the only team that can add that label.

PR #237, opened 2022-06-22
Bug 2100039: Adding support for disconnected clusters.

ybettan:
Disconnected clusters in OCP uses an ImageContentSourcePolicy resource
to map OCP image names to their mirror on a custom registry.

This ImageContentSourcePolicy writes mirroring configuration for the
container runtime on the nodes directly; the redirection is completely
transparent for other Kubernetes components, including SRO.

This means that the image referenced in the openshift/driver-toolkit
ImageStream cannot be directly pulled by SRO.

This PR is adding support for disconnected envs by wrapping crane
calls into a type that reads the system's registries.conf to resolve the
image name to the desired registry.

In addition, the wrapper sets up useful crane options:
  * optional auth, if it is configured in the pull secret
  * new transport with the pod's and the host's CA certificates
  * the current reconciliation's context

Co-authored-by: Quentin Barrand <quba@redhat.com>


ybettan:
/hold
Need to do some cleanup

PR #239, opened 2022-06-28
Read the openshift-config/user-ca-bundle ConfigMap

qbarrand:
This ConfigMap is configured at installation time and can contain a PEM-formatted bundle of CA certificates that should be copied on the node.
For SRO to be able to pull "just like the nodes' Docker daemon", we need to read that bundle (if it exists).

This PR extracts all the logic that fetches OpenShift user-defined CA certificates into a new `OpenShiftCAGetter` interface in a separate file.
It addresses [BZ 2100891](https://bugzilla.redhat.com/show_bug.cgi?id=2100891).

/cc @ybettan @yevgeny-shnaidman @cdvultur


ybettan:
Can you please have a commit message as well? `Git blame` will be pretty poor in the current state.

pkg/registry/openshift_ca.go
```diff
@@ -0,0 +1,87 @@
...
+
+type openShiftCAGetter struct {
+	kubeClient clients.ClientsInterface
+}
+
+//go:generate mockgen -source=openshift_ca.go -package=registry -destination=mock_openshift_ca_api.go
+
+type OpenShiftCAGetter interface {
+	AdditionalTrustedCAs(ctx context.Context) (map[string][]byte, error)
+	CABundle(context.Context) ([]byte, error)
```
ybettan, line 21:
```suggestion
	CABundle(ctx context.Context) ([]byte, error)
```
Just for consistency :)

qbarrand: Done.

pkg/registry/openshift_ca.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	return certs, nil
+
+}
+
+func (ocg *openShiftCAGetter) CABundle(ctx context.Context) ([]byte, error) {
+	logger := ctrl.LoggerFrom(ctx)
+
+	const userCABundleCMName = "user-ca-bundle"
```
ybettan, line 62: Although I like when const are defined just before there are use and in the smaller scope possible, all our consts are always defined at the top of each package.

Can we align with that?

ybettan: Then let's group all the const at the top of the function?

pkg/registry/wrapper.go
```diff
@@ -68,27 +71,40 @@ func (cw *craneWrapper) getPullSourcesForImageReference(imageName string) ([]sys
 }
 
 func (cw *craneWrapper) getClusterCustomCertPool(ctx context.Context) (*x509.CertPool, error) {
```
yevgeny-shnaidman, line 73: a general question: why are we extracting the CABundle and Cert on every reconciliation loop? Can we do it just once? Are we expecting the configmap in the cluster image to change?

qbarrand:
Looking into a better way to do this.

ybettan:
After talking to @yevgeny-shnaidman, there is an "Openshift way" to do it - see https://docs.openshift.com/container-platform/4.10/networking/configuring-a-custom-pki.html#certificate-injection-using-operators_configuring-a-custom-pki.

BUT, I don't think we should deviate much from 4.11 at this point. I think we should do it in KMMO and just add the reading of the additional confimap in the simplest way as possible in order to ease on the backport process.

Issue #228, opened 2022-06-15
SRO cannot deploy the simple-kmod example special resource


hershpa:
**Summary**
We install the SRO from **CLI** using the master branch of this repo. We see the following error in SRO controller manager logs: "Could not reconcile chart for SpecialResource","reconciler group":"sro.openshift.io","reconciler kind":"SpecialResource","name":"simple-kmod","namespace":"","error":"failed to get /simple-kmod runtime info: failed to get upgrade info: failed to extract dtk versions: DTK kernel not found running in the cluster. kernelFullVersion: 4.18.0-305.45.1.el8_4.x86_64. rtKernelFullVersion: 4.18.0-305.45.1.rt7.117.el8_4.x86_64"}"  We see the SRO is not deploying the simple-kmod example special resource due to above error. 

**Steps to reproduce:**
Env: OpenShift 4.10.16 Cluster
$ git clone https://github.com/openshift/special-resource-operator
$ cd special-resource-operator
$ TAG=master make deploy. We are specifically using **commit** 4e45f0dc74ce4c6383626001716eb6e64251df8e

The following will create "simple-kmod-0.0.1" ConfigMap in "default" namespace
$ ./scripts/make-cm-recipe ./charts/example/simple-kmod-0.0.1 simple-kmod-0.0.1 default

Also, there is a slight issue in Readme: 
Readme says "# name and namespace can be specified
$ ./scripts/make-cm-recipe ./charts/example/simple-kmod-0.0.1 namespace configmap-name"

However, configmap-name needs to be specified first followed by namespace. Like this:
"$ ./scripts/make-cm-recipe ./charts/example/simple-kmod-0.0.1 configmap-name namespace"

Finally: 
oc apply -f charts/example/simple-kmod-0.0.1/simple-kmod.yaml

The command runs successfully but no special resource created.

**Findings:**
We see the SRO is not deploying the simple-kmod example special resource. As a result, we looked at relevant logs and find below.
[...]

PR #240, opened 2022-06-29
Fixing the parameter order in scripts/make-cm-recipe.

ybettan:
Issue-228


ybettan:
Realtes to https://github.com/openshift/special-resource-operator/issues/228

Issue #243, opened 2022-07-13
Cannot build driver container on 4.10


Deixx:
I've been using SRO for some time to deploy kernel modules on a 4.10.14 cluster and it worked.
Today pod driver-build exited with error:
`error: build error: error determining starting point for build: no FROM statement found`

I suspect it has something to do with the last commit in `release-4.10` branch https://github.com/openshift/special-resource-operator/commit/e5dd074014c038ce4aea221c8d7c538ed9c0ed82, because I was able to deploy kmods on the same cluster without problems the previous time (about 2 weeks ago) and I've noticed some things in configs and logs:
- no value for IMAGE in BuildConfig:
```
  strategy:
    dockerStrategy:
      buildArgs:
      - name: IMAGE        # should have a value: {{ .Values.driverToolkitImage  }}
      - name: KMODVER
        value: v22.3
      - name: KVER
        value: 4.18.0-305.45.1.el8_4.x86_64
      dockerfilePath: Dockerfile.SRO
    type: Docker
```
- logs for pod special-resource-controller-manager have entries like:
```
INFO    warning         OnError: cannot extract manifest: could not build crane options: could not find a pull source for 
"quay.io/openshift-release-dev/ocp-release@sha256:e6aa643f494f64eefc8a6ad92b64156290bccd7e5cda56b8809f3a67ee2a53bb": 
could not find registry for image "quay.io/openshift-release-dev/ocp-
release@sha256:e6aa643f494f64eefc8a6ad92b64156290bccd7e5cda56b8809f3a67ee2a53bb": loading registries configuration 
"/mnt/host/registries.conf": open /mnt/host/registries.conf: no such file or directory
```
`INFO    open-cas-kmod   Runtime Information     {"DriverToolkitImage": ""}`
`INFO    open-cas-kmod   KernelAffine: ClusterUpgradeInfo        {"kernel": "4.18.0-305.45.1.el8_4.x86_64", "os": "8.4", "cluster": "4.10",
"driverToolkitImage": ""}`

Is it a bug or do I need to configure something additional to make this work?

PR #246, opened 2022-07-15
Bug 2107678: Update bundle (MGMT-11164)

enriquebelarte:
Bundle was not updated so builds could not mount /mnt/host/registries.conf as described in https://github.com/openshift/special-resource-operator/issues/243 and driver container build would end in an error.

### DIFF
--- a/bundle/4.10/manifests/openshift-special-resource-operator.v4.10.0.clusterserviceversion.yaml
+++ b/bundle/4.10/manifests/openshift-special-resource-operator.v4.10.0.clusterserviceversion.yaml
@@ -52,7 +52,7 @@ metadata:
       hardware accelerators or software defined storage, where extra management or
       kernel modules are needed.
     olm.skipRange: '>=4.6.0 <4.9.0'
-    operators.operatorframework.io/builder: operator-sdk-v1.15.0+git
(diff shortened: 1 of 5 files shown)


bundle/4.10/manifests/openshift-special-resource-operator.v4.10.0.clusterserviceversion.yaml
```diff
@@ -1304,7 +1312,7 @@ spec:
                   value: 0.0.1-snapshot
                 - name: SSL_CERT_DIR
                   value: /etc/pki/tls/certs
-                image: quay.io/openshift/origin-special-resource-rhel8-operator:4.10
+                image: quay.io/openshift-psap/special-resource-operator:release-4.10
```
qbarrand, line 1315: I don't think we should be using the PSAP image. I think you need to add `IMG=quay.io/openshift/origin-special-resource-rhel8-operator:4.10` to the `make` command you used to generate the bundle. cc @pacevedom

enriquebelarte: Done. Thank you.

bthurber (approved):
/lgtm

